PR #13, opened 2019-02-19
Adding ScalaStyle

rabee333:


scalastyle_config.xml
```diff
@@ -0,0 +1,188 @@
...
+ <check class="org.scalastyle.file.WhitespaceEndOfLineChecker" level="warning" enabled="true">
+   <parameters>
+    <parameter name="ignoreWhitespaceLines"><![CDATA[false]]></parameter>
+   </parameters>
+ </check>
+ <check class="org.scalastyle.scalariform.SpacesBeforePlusChecker" level="warning" enabled="true"></check>
+ <check class="org.scalastyle.file.FileLineLengthChecker" level="warning" enabled="true">
+  <parameters>
+   <parameter name="maxLineLength"><![CDATA[160]]></parameter>
+   <parameter name="tabSize"><![CDATA[4]]></parameter>
```
michazag: In IntelliJ scala-editor settings tab size defaults to '2'.
Let's change to '2', let me know if you think this is wrong

scalastyle_config.xml
```diff
@@ -0,0 +1,188 @@
...
+   <parameter name="ignoreImports"><![CDATA[false]]></parameter>
+  </parameters>
+ </check>
+ <check class="org.scalastyle.scalariform.ClassNamesChecker" level="warning" enabled="true">
+  <parameters>
+   <parameter name="regex"><![CDATA[^[A-Z][A-Za-z]*$]]></parameter>
```
michazag: Underscore ('_') should also be OK, here and for other names

scalastyle_config.xml
```diff
@@ -0,0 +1,188 @@
...
+  <parameters>
+   <parameter name="maxParameters"><![CDATA[8]]></parameter>
+  </parameters>
+ </check>
+ <check class="org.scalastyle.scalariform.MagicNumberChecker" level="warning" enabled="true">
+  <parameters>
+   <parameter name="ignore"><![CDATA[-1,0,1,2,3]]></parameter>
+  </parameters>
+ </check>
+ <check class="org.scalastyle.scalariform.NoWhitespaceBeforeLeftBracketChecker" level="warning" enabled="true"></check>
```
michazag: Is it typically expected to have whitespace before left bracket in every case in Scala?
If this is not commonly excepted we can omit

scalastyle_config.xml
```diff
@@ -0,0 +1,188 @@
...
+ <check class="org.scalastyle.scalariform.CovariantEqualsChecker" level="warning" enabled="true"></check>
+ <check class="org.scalastyle.scalariform.StructuralTypeChecker" level="warning" enabled="true"></check>
+ <check class="org.scalastyle.file.RegexChecker" level="warning" enabled="true">
+  <parameters>
+   <parameter name="regex"><![CDATA[println]]></parameter>
+  </parameters>
+ </check>
+ <check class="org.scalastyle.scalariform.NumberOfTypesChecker" level="warning" enabled="true">
+  <parameters>
+   <parameter name="maxTypes"><![CDATA[30]]></parameter>
```
michazag: What does this limit mean?

scalastyle_config.xml
```diff
@@ -0,0 +1,188 @@
...
+   <parameter name="regex"><![CDATA[^[a-z][A-Za-z]*$]]></parameter>
+   <parameter name="objectFieldRegex"><![CDATA[^[A-Z][A-Za-z]*$]]></parameter>
+  </parameters>
+ </check>
+ <check class="org.scalastyle.scalariform.TodoCommentChecker" level="warning" enabled="true">
+  <parameters>
+   <parameter name="words"><![CDATA[TODO|FIXME]]></parameter>
+  </parameters>
+ </check>
+</scalastyle>
```
michazag: GH complains about no newline at the end of the file.  I think we have a checker for that :)

scalastyle_config.xml
rabee333: Currently there are no classes with "_", and I think in Scala it's less recommended to have "_" in class names

scalastyle_config.xml
rabee333: This checks that there are not too many types declared in a file.

scalastyle_config.xml
rabee333: this is only before "[". 
I think it makes sense.

scalastyle_config.xml
rabee333: there is such a check at line 127:
 <check class="org.scalastyle.file.NewLineAtEofChecker" level="warning" enabled="true"></check>

rabee333:
@michazag 
I fixed the tab size.
As we discussed, this is the default scalastyle configuration. For now I only changed the tab size. Other changes can be made as needed.

PR #19, opened 2019-02-26
Add an option to override long running operation timeout limit

michazag:


michazag:
I'll close and re-issue this one after adjusting to the new folder structure

Issue #15, opened 2019-02-19
Add PrunedFilteredScan to read connector


michazag:
Extend KustoRelation with PrunedFilteredScan.

PR #23, opened 2019-02-28
Add pruning and filtering to the read-connector

michazag:
And also few minor fixes

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoFilter.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoFilter.scala
@@ -61,11 +61,10 @@ private[kusto] object KustoFilter {
   }
 
   private def binaryLogicalOperatorFilter(schema: StructType, leftFilter: Filter, rightFilter: Filter, operator: String): Option[String] = {
-    val left = buildFilterExpression(schema, leftFilter)
-    val right = if(left.isEmpty) None else buildFilterExpression(schema, rightFilter)
-
-    if (left.isEmpty || right.isEmpty) None else {
-      Some(s"(${left.get}) $operator (${right.get})")
+    buildFilterExpression(schema, leftFilter).flatMap { left =>
+      buildFilterExpression(schema, rightFilter).map { right =>
+        s"($left) $operator ($right)"
+      }
     }
   }
 
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoFilter.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoFilter.scala
@@ -1,0 +1,122 @@
+package com.microsoft.kusto.spark.datasource
+
+import java.sql.{Date, Timestamp}
+
+import org.apache.spark.sql.sources._
+import org.apache.spark.sql.types._
+
+private[kusto] object KustoFilter {
+  // Augment the original query to include column pruning and filtering
+  def pruneAndFilter(originalSchema: StructType, originalQuery: String, filtering: KustoFiltering): String = {
+    var query = originalQuery
+
+    if (!filtering.filters.isEmpty) {
+      query += KustoFilter.buildFiltersClause(originalSchema, filtering.filters)
+    }
+
+    if (!filtering.columns.isEmpty) {
+      query += KustoFilter.buildColumnsClause(filtering.columns)
+    }
+
+    query
+  }
+
+  def pruneSchema(schema: StructType, columns: Array[String]): StructType = {
+    val fieldMap = Map(schema.fields.map(x => x.name -> x): _*)
+    new StructType(columns.map(name => fieldMap(name)))
+  }
+
+  def buildColumnsClause(columns: Array[String]): String = {
+    if(columns.isEmpty) "" else {
+      " | project " + columns.mkString(", ")
+    }
+  }
+
+  def buildFiltersClause(schema: StructType, filters: Seq[Filter]): String = {
+    val filterExpressions = filters.flatMap(f => buildFilterExpression(schema, f)).mkString(" and ")
+    if (filterExpressions.isEmpty) "" else " | where " + filterExpressions
+  }
+
+  def buildFilterExpression(schema: StructType, filter: Filter): Option[String] = {
+
+    filter match {
+      case EqualTo(attr, value) => binaryScalarOperatorFilter(schema, attr, value, "==")
+      case EqualNullSafe(attr, value) if value == null => unaryScalarOperatorFilter(attr, "isnull")
+      case EqualNullSafe(attr, value) => binaryScalarOperatorFilter(schema, attr, value, "==")
+      case GreaterThan(attr, value) => binaryScalarOperatorFilter(schema, attr, value, ">")
+      case GreaterThanOrEqual(attr, value) => binaryScalarOperatorFilter(schema, attr, value, ">=")
+      case LessThan(attr, value) => binaryScalarOperatorFilter(schema, attr, value, "<")
+      case LessThanOrEqual(attr, value) => binaryScalarOperatorFilter(schema, attr, value, "<=")
+      case In(attr, values) => unaryOperatorOnValueSetFilter(schema, attr, values, "in")
+      case IsNull(attr) => unaryScalarOperatorFilter(attr, "isnull")
+      case IsNotNull(attr) => unaryScalarOperatorFilter(attr, "isnotnull")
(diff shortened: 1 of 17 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoFilter.scala
```diff
@@ -0,0 +1,122 @@
...
+  private  def stringOperatorFilter(schema: StructType, attr: String, value: String, operator: String): Option[String] = {
+    // Will return 'None' if 'attr' is not part of the 'schema'
+    getType(schema, attr).map {
+      _ => s"""$attr $operator '$value'"""
+    }
+  }
+
+  private def toStringList(values: Array[Any], dataType: DataType): String = {
+    val combined = values.map(value => format(value, dataType)).mkString(", ")
+    if (combined.isEmpty) "" else combined
```
LizaShak: I think if values is empty then the mkString will return just an empty string, so you don't need this check and you can just write: values.map(value => format(value, dataType)).mkString(", ")

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoFilter.scala
```diff
@@ -0,0 +1,122 @@
...
+    Some(s"$function($attr)")
+  }
+
+  private def binaryLogicalOperatorFilter(schema: StructType, leftFilter: Filter, rightFilter: Filter, operator: String): Option[String] = {
+    val left = buildFilterExpression(schema, leftFilter)
+    val right = if(left.isEmpty) None else buildFilterExpression(schema, rightFilter)
+
+    if (left.isEmpty || right.isEmpty) None else {
+      Some(s"(${left.get}) $operator (${right.get})")
+    }
```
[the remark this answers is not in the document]
michazag: Nice. But where does the $operator fit here?

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoFilter.scala
```diff
@@ -0,0 +1,122 @@
...
+      case LessThanOrEqual(attr, value) => binaryScalarOperatorFilter(schema, attr, value, "<=")
+      case In(attr, values) => unaryOperatorOnValueSetFilter(schema, attr, values, "in")
+      case IsNull(attr) => unaryScalarOperatorFilter(attr, "isnull")
+      case IsNotNull(attr) => unaryScalarOperatorFilter(attr, "isnotnull")
+      case And(left, right) => binaryLogicalOperatorFilter(schema, left, right, "and")
+      case Or(left, right) => binaryLogicalOperatorFilter(schema, left, right, "or")
+      case Not(child) => unaryLogicalOperatorFilter(schema, child, "not")
+      case StringStartsWith(attr, value) => stringOperatorFilter(schema, attr, value, "startswith_cs")
+      case StringEndsWith(attr, value) => stringOperatorFilter(schema, attr, value, "endswith_cs")
+      case StringContains(attr, value) => stringOperatorFilter(schema, attr, value, "contains_cs")
```
ohadbitt: we should consider using 'has' which is much much faster but is only for whole terms capturing

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoFilter.scala
```diff
@@ -0,0 +1,122 @@
...
+
+    if (!filtering.columns.isEmpty) {
+      query += KustoFilter.buildColumnsClause(filtering.columns)
+    }
+
+    query
+  }
+
+  def pruneSchema(schema: StructType, columns: Array[String]): StructType = {
+    val fieldMap = Map(schema.fields.map(x => x.name -> x): _*)
```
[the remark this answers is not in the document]
michazag: Much better 👍

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoFilter.scala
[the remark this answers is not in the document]
ohadbitt: I think u confused here a little - as this will drop the left filter ..

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoFilter.scala
```diff
@@ -0,0 +1,122 @@
...
+      Some(s"(${left.get}) $operator (${right.get})")
+    }
+  }
+
+  private def unaryLogicalOperatorFilter(schema: StructType, childFilter: Filter, operator: String): Option[String] = {
+    val child = buildFilterExpression(schema, childFilter)
+
+    if (child.isEmpty) None else {
+      Some(s"$operator(${child.get})")
+    }
```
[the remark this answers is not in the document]
ohadbitt: maybe
buildFilterExpression(schema, childFilter).map(c => s"$operator($c)")

connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
```diff
@@ -5,19 +5,19 @@ import com.microsoft.kusto.spark.datasink.KustoWriter.TempIngestionTablePrefix
...
     s""".show database $database extents metadata | where TableName startswith '$TempIngestionTablePrefix' | project TableName, maxDate = todynamic(ExtentMetadata).MaxDateTime | where maxDate > ago(1h)"""
   }
 
-  def generateFindOldTempTablesCommand(database: String): String = {
+  private[kusto] def generateFindOldTempTablesCommand(database: String): String = {
     s""".show journal | where Event == 'CREATE-TABLE' | where Database == '$database' | where EntityName startswith '$TempIngestionTablePrefix' | where EventTimestamp < ago(1h) and EventTimestamp > ago(3d) | project EntityName """
   }
 
-  def generateFindCurrentTempTablesCommand(prefix: String): String = {
+  private[kusto] def generateFindCurrentTempTablesCommand(prefix: String): String = {
```
ohadbitt: why?
this can be open for users who want to run queries on java client ..
and anyway why did u leave some as public ?

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoFilter.scala
ohadbitt: we can also consider some configuration settings for the case sensitivity

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoFilter.scala
```diff
@@ -0,0 +1,122 @@
...
+    if (!filtering.columns.isEmpty) {
+      query += KustoFilter.buildColumnsClause(filtering.columns)
+    }
+
+    query
+  }
+
+  def pruneSchema(schema: StructType, columns: Array[String]): StructType = {
+    val fieldMap = Map(schema.fields.map(x => x.name -> x): _*)
+    new StructType(columns.map(name => fieldMap(name)))
```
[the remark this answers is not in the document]
michazag: It is possible.
So, should I revert to Map(schema.fields.map(x => x.name -> x): _*) ?

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoFilter.scala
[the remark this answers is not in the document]
LizaShak: You are right, sorry, I have ignored the last two lines, I meant something like that:

buildFilterExpression(schema, leftFilter).flatMap { left =>
      buildFilterExpression(schema, rightFilter).map { right =>
        s"(${left}) $operator (${right})"
      }
    }

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoFilter.scala
michazag: "has" - since it is really a string comparison (and not necessarily a whole entity) I don't see how I can use it. 
Case sensitivity - I'll add a task, maybe we'll add as a configuration option

connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
michazag: I don't think we should allow Java client dependencies on the connector.
On the second - missed entries added in the merge, will fix. thanks

michazag: Thinking again... I just made the whole object private[kusto]

michazag:
Closes #15

Issue #22, opened 2019-02-27
Improve performance when stream writing from Event Hub with more than one partition


navalev:
Current streaming test is not efficient when stream writing from an Event Hub with more than one partition.
Verified issue is not related to event hub spark connector  - tested writing to local file storage, read from event hub partitions with sufficient performance

navalev:
When running with *KUSTO_WRITE_ENABLE_ASYNC* set to *false* the performance with more than one partition is improved (skips foreachPartitionAsync in KustoWriter that is blocker with await)

PR #26, opened 2019-03-05
Minor changes to readme and spark streaming sample

navalev:
- Update readme beta info on maven dependency
 - update gitignore
 - update spark streaming sample to set async to false by default as a temporary workaround to improve performance #22


README.md
```diff
@@ -37,6 +37,8 @@ version = 1.0.0-Beta-01
 ```
 
 **In Maven**:
+
+Note that the jar is in beta and not available yet in public maven. clone this repository and build it localy to add it to your local maven repository, or use the [pre-built jars](lib/)
```
michazag: -> Note: (add the '->' is for .md formatting)

. clone - . Clone (capitalize)

samples/src/main/scala/SparkStreamingKustoSink.scala
```diff
@@ -47,7 +47,7 @@ object SparkStreamingKustoSink {
       .option(KustoOptions.KUSTO_TABLE, "Your Kusto Destination Table")
       .option(KustoOptions.KUSTO_AAD_CLIENT_ID, "Your Client ID")
       .option(KustoOptions.KUSTO_AAD_CLIENT_PASSWORD, "Your secret")
-      .option(KustoOptions.KUSTO_WRITE_ENABLE_ASYNC, "true")
+      .option(KustoOptions.KUSTO_WRITE_ENABLE_ASYNC, "false")
```
michazag: This is the default, so you can just omit this option.

README.md
navalev: done.

samples/src/main/scala/SparkStreamingKustoSink.scala
```diff
@@ -9,6 +9,10 @@ import org.apache.spark.sql.functions._
 /** ************************************************/
 /*          STREAMING SINK EXAMPLE                */
 /** ************************************************/
+
+// To enable faster ingestion into kusto, set a  minimal value for the batching ingestion policy:
```
ohadbitt: This should be followed with a warning for pricing and maybe even a link to the docs : https://kusto.azurewebsites.net/docs/concepts/batchingpolicy.html?q=batching

michazag: Would be better to point to the external documentation at https://docs.microsoft.com/en-us/azure/kusto/concepts/ 
The only problem is that the batching policy section was not published. Vladik made the necessary change already, I just don't see it reflected on the documentation site yet

PR #30, opened 2019-03-11
split large partitions

LizaShak:


connector/pom.xml
```diff
@@ -224,6 +224,11 @@
       <artifactId>azure-keyvault-webkey</artifactId>
       <version>1.1</version>
     </dependency>
+    <dependency>
+      <groupId>org.scalatest</groupId>
+      <artifactId>scalatest-funspec_2.11</artifactId>
```
michazag: See line 191 - duplicate?

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -41,144 +45,109 @@ object KustoWriter{
   val statusCol = "Status"
   val delayPeriodBetweenCalls: Int = KCONST.DefaultPeriodicSamplePeriod.toMillis.toInt
   val GZIP_BUFFER_SIZE: Int = 16 * 1024
+  val GIGA_SIZE: Int = 1024 * 1024 * 1024
```
michazag: Please move to KCONST

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -41,144 +45,109 @@ object KustoWriter{
...
-                           data: DataFrame,
-                           tableCoordinates: KustoCoordinates,
-                           authentication: KustoAuthentication,
-                           writeOptions: WriteOptions): Unit = {
-    val schema = data.schema
-    var batchIdIfExists = { if (batchId.isEmpty || batchId == Option(0)) { "" } else { s"${batchId.get}"} }
+  private[kusto] def write(batchId: Option[Long], data: DataFrame,
+                           tableCoordinates: KustoCoordinates, authentication: KustoAuthentication, writeOptions: WriteOptions): Unit = {
 
+    val batchIdIfExists = batchId.map(_.toString).getOrElse("")
```
michazag: Need to also account for batchId == Option(0) (handled same as if empty)

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -41,144 +45,109 @@ object KustoWriter{
...
-    }
-    catch  {
-      case ex: Exception =>
-        KDSU.reportExceptionAndThrow(myName, ex, "trying to drop temporary tables", tableCoordinates.cluster, tableCoordinates.database, table, isLogDontThrow = true)
-    }
-
-    val tmpTableName = KustoQueryUtils.simplifyName(s"$TempIngestionTablePrefix${appName}_$table${batchIdIfExists}_${UUID.randomUUID().toString}")
-
-    if(batchIdIfExists != "") batchIdIfExists = s", batch'$batchIdIfExists'"
+    cleanupTempTables(kustoAdminClient, tableCoordinates)
```
michazag: Please be careful here when merging with Ohad's change :)

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -189,94 +158,164 @@ object KustoWriter{
...
+
+
+    Await.result(
+      Future(ingestRowsIntoKusto(rows, storageUri)).map { ingestionResults =>
+        // Proceed only on success. Will throw on failure for the driver to handle
+        ingestionResults.foreach { ingestionResult =>
+          KDSU.runSequentially[IngestionStatus](
+            func = () => ingestionResult.GetIngestionStatusCollection().get(0),
+            0, delayPeriodBetweenCalls, (timeout.toMillis / delayPeriodBetweenCalls + 5).toInt,
+            res => res.status == OperationStatus.Pending,
```
michazag: How about the idea that we discussed to only pass result operation-ids to be aggregated by the driver, and the driver would test all results to completion (the driver also knows how many results to expect and can throw if the expected number is not matched within a few minutes)?

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -189,94 +158,164 @@ object KustoWriter{
...
+    while (it.hasNext && res) {
       res = it.next()
     }
     res
   }
 
-  @throws[IOException]
-  private[kusto] def serializeRows(rows: Iterator[InternalRow], schema: StructType, bos: BlobOutputStream, timeZone: String): Unit = {
-    val gzip = new GZIPOutputStream(bos)
+  def createFileWriter(schema: StructType,
```
michazag: blobWriter?

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -189,94 +158,164 @@ object KustoWriter{
...
-      csvWriter.writeRow(csvSerializer.convertRow(row))
+  @throws[IOException]
+  private[kusto] def serializeRows(rows: Iterator[InternalRow],
+                                   storageUri: String,
+                                   parameters: KustoWriteResource): Seq[CloudBlockBlob] = {
+    import parameters._
+    val container: CloudBlobContainer = new CloudBlobContainer(new URI(storageUri))
+    val fileWriter: FileWriteResource = createFileWriter(schema, writeOptions.timeZone, coordinates, tmpTableName, storageUri, container)
+
+    val (_, _, blobs) = rows.foldLeft[(Long, FileWriteResource, Seq[CloudBlockBlob])]((0, fileWriter, Seq())) { case ((size, fileWriter, blobsCreated), row) =>
```
michazag: Some comments could be helpful here

connector/src/test/scala/com/microsoft/kusto/spark/KustoWriterTests.scala
```diff
@@ -0,0 +1,59 @@
...
+import com.microsoft.kusto.spark.datasink.{FileWriteResource, KustoWriter}
+import com.univocity.parsers.csv.{CsvWriter, CsvWriterSettings}
+import org.apache.spark.SparkConf
+import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.apache.spark.sql.catalyst.InternalRow
+import org.mockito.Mockito._
+import org.scalatest.{FunSpec, Matchers}
+
+
+class KustoWriterTests extends FunSpec with Matchers {
```
michazag: We used FlatSpec in all the other tests. It would be better to have all tests aligned in style.

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
LizaShak: the first one is the size of the last row in the dataframe, and the second one is the last blob that have created, I am not using those parameters after that, so if I give them a name the scalastyle marks them as unused... I can still give them a name and make scalastyle ignore it, what do you think?

connector/pom.xml
LizaShak: It is for the funspec I used in my tests, I converted mine to flatSpec and removed that one

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
michazag: It's OK not to give names, just add a comment (the first one is the size of the last row in the data frame, and the second one is the last blob that was created)

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
LizaShak: Definetly. I will do it later

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
ohadbitt: it seems like an overkill for the foldleft function - as both variables fileWriter, blobsCreated dont change most of the operation . 
i think it will be better without the 'case'
and without declaring fileWriter before (u can simply return  (_ , fileWriter, blobs)  )

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -189,94 +158,164 @@ object KustoWriter{
...
 
-    resultAsJson
+    val (fields, size) = row.toSeq(schema).foldLeft[(Seq[String], Int)](Seq[String](), 0) { (res, curr) =>
+      val formattedField: String = curr match {
+        case DateType => DateTimeUtils.toJavaDate(row.getInt(res._1.size)).toString
+        case TimestampType => dateFormat.format(DateTimeUtils.toJavaTimestamp(row.getLong(res._1.size)))
+        case _ => row.get(res._1.size, schemaFields(res._1.size).dataType).toString
+
+      }
+      (res._1 :+ formattedField, res._2 + formattedField.getBytes(StandardCharsets.UTF_8).length)
```
ohadbitt: seems expensive - we should check performance changes

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -189,94 +158,164 @@ object KustoWriter{
...
+  def createFileWriter(schema: StructType,
+                       timeZone: String,
+                       tableCoordinates: KustoCoordinates,
+                       tmpTableName: String,
+                       storageUri: String,
+                       container: CloudBlobContainer): FileWriteResource = {
+    val blobName = s"${tableCoordinates.database}_${tmpTableName}_${UUID.randomUUID.toString}_SparkStreamUpload.gz"
+    val blob: CloudBlockBlob = container.getBlockBlobReference(blobName)
+    val gzip: GZIPOutputStream = new GZIPOutputStream(blob.openOutputStream())
     val csvSerializer = new KustoCsvSerializationUtils(schema, timeZone)
```
ohadbitt: can delete this and KustoCsvSerializationUtils previous implementation

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
LizaShak: I did some research, untill Scala 2.11 Seq had the same complexity like Vector in that manner. 
And now it has linear complexity for appending items. So after some research, **I think in that case the most efficient way will be to prepend a value to LIst**. 
Based on that article: [Performance Comparison between immutable Seq, List, Vector](https://danielasfregola.com/2015/06/15/which-immutable-scala-collection/)
And List/Seq/Vector data structure implementation.
What do you think?

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -189,94 +158,164 @@ object KustoWriter{
...
+      val formattedField: String = curr match {
+        case DateType => DateTimeUtils.toJavaDate(row.getInt(res._1.size)).toString
+        case TimestampType => dateFormat.format(DateTimeUtils.toJavaTimestamp(row.getLong(res._1.size)))
+        case _ => row.get(res._1.size, schemaFields(res._1.size).dataType).toString
+
+      }
+      (res._1 :+ formattedField, res._2 + formattedField.getBytes(StandardCharsets.UTF_8).length)
+    }
+
+    CsvRowResult(fields.toArray, size + fields.size)
```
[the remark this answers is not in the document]
LizaShak: Because of the commas you have in csv row, separating the fields.

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
LizaShak: I think it is a Scala idiom to prefer immutable code. 
I also think it prefers it because of the way it works behind scenes, so using val is usually more efficient (there are some cases that it is not, but not in ours I think). 
[link1](https://www.quora.com/Scala-Are-there-any-performance-implications-to-using-vars-versus-vals)
[link2](https://en.wikipedia.org/wiki/Scala_(programming_language)#Immutability)
So yes fileWriter can be a var and then we can update it later... But I don't see the need. 
The foldLeft has to start from something and it accumulates the values. This is why I define the fileWriter before that. I don't want to construct it every iteration, just when I need to. I am also using the case so I can give the tuple fields names to make  it more readable. I don't think it is less efficient or something, there are no types specified so I don't think it does like an extra check or something in that case. But I am not sure and didn't find any evidence to support that, just an intuition then...

ohadbitt: ok, your call. 
anyway, I  wasnt talking about it being var . The first line can simply be : 
    val (_, fileWriter, blobs) = rows.foldLeft[(Long, FileWriteResource, Seq[CloudBlockBlob])]((0, createFileWriter(schema, writeOptions.timeZone, coordinates, tmpTableName, storageUri, container), Seq())) { case ((size, fileWriter, blobsCreated), row) => ....

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
ohadbitt: sorry for not being clear in my comments 
i was actually referring to the getBytes method . 
But it was an interesting read - in that manner it seems you are right

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
LizaShak: ahhh I need it to be declared first also for finalizing it later in case the partition size was less than a GB

michazag: I think its also more readable when created first... in any case adding some comments for future generations can be helpful here

PR #48, opened 2019-04-15
Parse table schema in Kusto

michazag:
Parse table schema in kusto, and get the result as rows of name/csl-type/type.
This is both more efficient and reduces dependency on Jackson-core library.
Also, for query-schema it is sufficient to take 0 (instead of take 1).
All changes are pending validation

### DIFF
--- a/connector/pom.xml
+++ b/connector/pom.xml
@@ -79,11 +79,6 @@
             <version>1.6.5</version>
         </dependency>
         <dependency>
-            <groupId>org.json4s</groupId>
-            <artifactId>json4s-jackson_${scala.version}</artifactId>
-            <version>3.5.3</version>
-        </dependency>
-        <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
             <version>2.8.11.2</version>
--- a/connector/src/test/scala/com/microsoft/kusto/spark/KustoWriterTests.scala
+++ b/connector/src/test/scala/com/microsoft/kusto/spark/KustoWriterTests.scala
@@ -2,9 +2,11 @@ package com.microsoft.kusto.spark
 
 import java.io.BufferedWriter
 import java.nio.charset.StandardCharsets
+import java.util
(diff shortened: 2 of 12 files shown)


connector/src/test/scala/com/microsoft/kusto/spark/KustoWriterTests.scala
```diff
@@ -52,4 +54,31 @@ class KustoWriterTests extends FlatSpec with Matchers {
     verify(buffer, times(1)).close()
   }
 
+  "getColumnsSchema" should "parse table schema correctly" in {
+    //val columns = "{\"Name\":\"Subscriptions\",\"OrderedColumns\":[{\"Name\":\"SubscriptionGuid\",\"Type\":\"System.String\",\"CslType\":\"string\"},{\"Name\":\"Identifier\",\"Type\":\"System.String\",\"CslType\":\"string\"},{\"Name\":\"SomeNumber\",\"Type\":\"System.Int64\",\"CslType\":\"long\"},{\"Name\":\"IsCurrent\",\"Type\":\"System.SByte\",\"CslType\":\"bool\"},{\"Name\":\"LastModifiedOn\",\"Type\":\"System.DateTime\",\"CslType\":\"datetime\"},{\"Name\":\"IntegerValue\",\"Type\":\"System.Int32\",\"CslType\":\"int\"}]}"
```
ohadbitt: delete?

connector/pom.xml
```diff
@@ -78,11 +78,6 @@
             <artifactId>azure-client-runtime</artifactId>
             <version>1.6.5</version>
         </dependency>
-        <dependency>
```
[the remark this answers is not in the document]
michazag: :)

connector/src/test/scala/com/microsoft/kusto/spark/KustoWriterTests.scala
michazag: I change a bit but left since this explains the compared schema

PR #55, opened 2019-05-12
Support nested objects serialization

toshetah:
So far we tried to use the row.get() function to extract its value, which gave encoded strings, instead of their real values.
This PR tries to extract the data recursively.


connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -337,6 +330,71 @@ object KustoWriter {
     CsvRowResult(fields.toArray, size + fields.size)
   }
 
+  private def convertOneField(row: InternalRow, schemaFields: Array[StructField], fieldIndexInRow: Int, timeZone: String) = {
+    val dateFormat = FastDateFormat.getInstance("yyyy-MM-dd'T'HH:mm:ss.SSSXXX", TimeZone.getTimeZone(timeZone))
```
danield137: this seems redundant. can't this happen once somewhere in higher scope?

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -337,6 +330,71 @@ object KustoWriter {
     CsvRowResult(fields.toArray, size + fields.size)
   }
 
+  private def convertOneField(row: InternalRow, schemaFields: Array[StructField], fieldIndexInRow: Int, timeZone: String) = {
```
danield137: `convertField`

danield137: reorder params. 
`row: InternalRow, fieldIndexInRow: Int, schemaFields: Array[StructField], , timeZone: String`

danield137: also, isn't it kinda redundant to pass the entire schema fields array? you know what field you are operating on...

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -337,6 +330,71 @@ object KustoWriter {
...
+  }
+
+  private def convertStructToCsv(row: InternalRow, schema: StructType, timeZone: String): String = {
+    val dateFormat = FastDateFormat.getInstance("yyyy-MM-dd'T'HH:mm:ss.SSSXXX", TimeZone.getTimeZone(timeZone))
+    val fields = schema.fields
+    val result:scala.collection.mutable.Map[String,String] = scala.collection.mutable.Map()
+
+    for (x <- 0 to fields.length - 1) {
+      val dataType = fields(x).dataType
+      val value = if (row.get(x, fields(x).dataType) == null) null else dataType match {
```
danield137: should this recursively call `convertOneField`?

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -337,6 +330,71 @@ object KustoWriter {
...
+      case TimestampType => dateFormat.format(DateTimeUtils.toJavaTimestamp(row.getLong(fieldIndexInRow)))
+      case StringType => row.getString(fieldIndexInRow)
+      case BooleanType => row.getBoolean(fieldIndexInRow).toString
+      case _ : StructType => convertStructToCsv(row.getStruct(fieldIndexInRow, dataType.asInstanceOf[StructType].length), dataType.asInstanceOf[StructType], timeZone)
+      case _ : ArrayType => ConvertArrayToCsv(row.getArray(fieldIndexInRow), dataType, timeZone)
+      case _ => row.get(fieldIndexInRow, dataType).toString
+    }
+  }
+
+  private def convertStructToCsv(row: InternalRow, schema: StructType, timeZone: String): String = {
```
danield137: why is this called `convertStructToCsv` whereas `convertOneField` is missing the csv postfix?

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -337,6 +330,71 @@ object KustoWriter {
...
+
+      if (value != null) {
+        result(fields(x).name) = value
+      }
+    }
+
+    "{" + result.map{case (k, v) => "\"" +  k + "\"" + ":" + v }.mkString(",") + "}"
+  }
+
+  private def ConvertArrayToCsv(ar: ArrayData, fieldsType: DataType, timeZone: String): String = {
```
danield137:
```suggestion
  private def convertArrayToCsv(ar: ArrayData, fieldsType: DataType, timeZone: String): String = {
```

lower case

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -337,6 +330,71 @@ object KustoWriter {
...
+    val dateFormat = FastDateFormat.getInstance("yyyy-MM-dd'T'HH:mm:ss.SSSXXX", TimeZone.getTimeZone(timeZone))
+    if (ar == null) null else {
+      if (ar.numElements() == 0) "[]" else {
+        val result:Array[String] = new Array(ar.numElements())
+        val fieldType = fieldsType.asInstanceOf[ArrayType].elementType
+        for (x <- 0 to ar.numElements() - 1) {
+          result(x) = if (ar.get(x, fieldType) == null) "" else fieldType match {
```
danield137: again, shouldn't it be `convertOneField`?

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
danield137: I might not know scala enough, but shouldn't this also have a return type?

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
toshetah: No, because the parsing is a bit different between the convert functions.
For example the quotes around the string in the convertStructToCsv method

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
toshetah: convertOneField no longer exists.

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
ohadbitt: why do you need to call row.get here - when you call the explicit "get" typed function later

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -337,6 +330,71 @@ object KustoWriter {
...
+          }
+        }
+
+        "[" + result.mkString(",") + "]"
+      }
+    }
+  }
+
+  private def GetStringFromArray(ar: ArrayData, x: Int) = {
+    if (ar.getUTF8String(x) == null) "" else ar.getUTF8String(x).toString
```
ohadbitt: if ar.getUTF8String(x)  is not null you calculate it twice..

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -11,10 +11,10 @@ import java.util.zip.GZIPOutputStream
 import java.util.{TimeZone, UUID}
```
ohadbitt: reformat file

.gitignore
```diff
@@ -24,6 +24,8 @@ out/
 # JDT-specific (Eclipse Java Development Tools)
 .classpath
 scalastyle-output.xml
+*.prefs
+*.factorypath
```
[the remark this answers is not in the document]
toshetah: probably intelij files.

connector/src/test/scala/com/microsoft/kusto/spark/KustoWriterTests.scala
```diff
@@ -3,40 +3,60 @@ package com.microsoft.kusto.spark
...
 import org.mockito.Mockito._
+import org.scalatest.junit.JUnitRunner
 import org.scalatest.{FlatSpec, Matchers}
 
 
+@RunWith(classOf[JUnitRunner])
 class KustoWriterTests extends FlatSpec with Matchers {
 
-  def getDF(): DataFrame = {
+  def getDF(short: Boolean): DataFrame = {
```
ohadbitt: i think a name like - "isNestedSchema" will help understand the differences

connector/src/test/resources/TestData/TestDynamicFields.csv
```diff
@@ -0,0 +1 @@
+﻿{"string":"value","int_ar": [1,2,3],"string_ar": ["a","b","c"],"bool_ar": [true,false],"string_ar_ar":[["a","b","c"]],"struct_struct":{"bool":true,"int":1,"string":"abc","string_ar":["a","b","c"],"int_ar":[1,2,3],"dict_ar":[{"string":"a","int":1},{"string":"b","int":2}]}}
```
ohadbitt: txt file ?

PR #67, opened 2019-05-28
Consolidate Kusto clients

toshetah:
This is a the first change in a series that will consolidate Kusto logic in the same classes, instead of splitting them in different files.


connector/src/main/scala/com/microsoft/kusto/spark/authentication/KustoAuthentication.scala
```diff
@@ -0,0 +1,52 @@
...
+  def canEqual(that: Any) : Boolean = that.isInstanceOf[AadApplicationAuthentication]
+  override def equals(that: Any) : Boolean = that match {
+    case auth : KeyVaultCertificateAuthentication => uri == auth.uri && pemFilePath == auth.pemFilePath
+    case _ => false
+  }
+
+  override def hashCode(): Int = uri.hashCode + pemFilePath.hashCode
+}
+
+case class KustoAccessTokenAuthentication(token: String) extends KustoAuthentication {
```
danield137: What is access token? is that a validated auth token? if so, shouldn't this be just `KustoTokenAuthentication`?

connector/src/main/scala/com/microsoft/kusto/spark/authentication/KustoAuthentication.scala
```diff
@@ -0,0 +1,52 @@
+package com.microsoft.kusto.spark.authentication
```
danield137: shouldn't this be implemented in our Java client?

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
```diff
@@ -1,59 +1,82 @@
...
+import org.joda.time.{DateTime, DateTimeZone, Period}
 
-  def getIngest(authentication: KustoAuthentication, clusterAlias: String): IngestClient = {
-    val ingestKcsb = getKcsb(authentication, s"https://ingest-$clusterAlias.kusto.windows.net")
-    ingestKcsb.setClientVersionForTracing(KCONST.clientName)
-    getIngest(ingestKcsb, clusterAlias)
-  }
+import scala.collection.JavaConverters._
+
+class KustoClient(val clusterAlias: String, val engineClient: Client, val dmClient: Client, val ingestClient: IngestClient) {
```
danield137: I would avoid calling this class `KustoClient` to prevent ambiguity. 
A better name would be `KustoBackend` IMHO, or just `KustoDatabase` as it has some database related logic

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClientCache.scala
```diff
@@ -0,0 +1,65 @@
+package com.microsoft.kusto.spark.utils
+
+import java.util.concurrent.ConcurrentHashMap
+
+import com.microsoft.azure.kusto.data.{ClientFactory, ConnectionStringBuilder}
+import com.microsoft.azure.kusto.ingest.IngestClientFactory
+import com.microsoft.kusto.spark.authentication.{AadApplicationAuthentication, KeyVaultAuthentication, KustoAccessTokenAuthentication, KustoAuthentication}
+import com.microsoft.kusto.spark.utils.{KustoConstants => KCONST}
+
+object KustoClientCache {
```
danield137: I wouldn't call this a cache. it it more of a factory or a pool than a cache (caching is just an implementation detail). 
`KustoBackendFactory` or `KustoDatabasePool`

connector/src/test/scala/com/microsoft/kusto/spark/authentication/KustoAuthenticationTests.scala
```diff
@@ -0,0 +1,59 @@
+package com.microsoft.kusto.spark.authentication
+
+import org.junit.runner.RunWith
+import org.scalatest.FlatSpec
+import org.scalatest.junit.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class kustoAuthenticationTests extends FlatSpec {
+  "KeyVaultAppAuthentication Equals" should "Check equality and inequality between different KeyVaultAppAuthentication" in {
```
danield137: definitely should be part of sdk. unrelated to spark.

connector/src/main/scala/com/microsoft/kusto/spark/authentication/KustoAuthentication.scala
toshetah: This change is only cosmetic. So let's think about it later.

connector/src/main/scala/com/microsoft/kusto/spark/authentication/KustoAuthentication.scala
toshetah: I am not sure. Same as above.

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
toshetah: We'll think about it. Same as above.

connector/src/test/scala/com/microsoft/kusto/spark/authentication/KustoAuthenticationTests.scala
toshetah: One day :)

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClientCache.scala
toshetah: Once we change KustoClient, we need to update this as well.

PR #74, opened 2019-06-16
poll on driver

ohadbitt:
poll on driver


connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -92,23 +95,22 @@ object KustoWriter {
     }
     else {
       try {
-        rdd.foreachPartition { rows => ingestRowsIntoTempTbl(rows, batchIdIfExists, writeOptions.timeout, storageUri) }
+        rdd.foreachPartition { rows => ingestRowsIntoTempTbl(rows, batchIdIfExists, storageUri, acc) }
```
toshetah: It's better not to pass the storageUri.
You should pass a KustoAuthentication, and then each worker can resolve its own storage.
We will get more spread data between the storage accounts.

toshetah: We can do that in another pull request, if you'd like.

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -192,66 +195,75 @@ object KustoWriter {
...
-        ingestionResults.foreach {
-          ingestionResult =>
+    val partitionsResults = acc.value
+    val mergeTask = Future {
+      try {
+        partitionsResults.asScala.foreach {
+          partitionResult =>
             KDSU.runSequentially[IngestionStatus](
-              func = () => ingestionResult.getIngestionStatusCollection.get(0),
+              func = () => partitionResult.ingestionResult.getIngestionStatusCollection.get(0),
```
toshetah: This logic should be at KustoClient.
Each table call is independent of the others, therefore I don't see any value of merging them all together.
Let's expose at KustoCLient a method called: waitForResultToComplete, which will return Future, and then we can wait on all of those.

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -6,20 +6,18 @@ import java.net.URI
...
-import com.microsoft.azure.kusto.ingest.result.{IngestionResult, IngestionStatus, OperationStatus}
+import com.microsoft.azure.kusto.ingest.result.IngestionResult
 import com.microsoft.azure.kusto.ingest.source.BlobSourceInfo
 import com.microsoft.azure.kusto.ingest.{IngestClient, IngestionProperties}
 import com.microsoft.azure.storage.StorageCredentialsSharedAccessSignature
 import com.microsoft.azure.storage.blob.{CloudBlobContainer, CloudBlockBlob}
 import com.microsoft.kusto.spark.authentication.KustoAuthentication
-import com.microsoft.kusto.spark.datasink
 import com.microsoft.kusto.spark.common.KustoCoordinates
+import com.microsoft.kusto.spark.datasink
```
danield137: not sure what tool you are using , but might want to reorder imports ?

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
```diff
@@ -79,4 +87,79 @@ class KustoClient(val clusterAlias: String, val engineClient: Client, val dmClie
...
+      }
+    }
+
+    if(!isAsync)
+    {
+      Await.result(mergeTask, timeout)
+    }
+  }
+
+  private[kusto] def tryCleanupIngestionByproducts(database: String, kustoAdminClient: Client, tmpTableName: String): Unit = {
```
danield137: In my experience, try usually means it can fail. would consider renaming
https://stackoverflow.com/q/27459937/929164

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
ohadbitt: intellij 
funny that this happened

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
ohadbitt: your experience is right - this can fail - and what we mean here that we dont care if it failed or not

danield137: if you don't care about if it can fail, then remove the try in the name. 
when you read code with `try` prefix, you expect it to be wrapped with try catch..

PR #77, opened 2019-06-19
DontFailOnCountTimeout

ohadbitt:
Dont fail on count timeout

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
@@ -11,8 +11,12 @@ import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.sources.{BaseRelation, Filter, PrunedFilteredScan, TableScan}
 import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.{Row, SQLContext, SparkSession}
+import com.microsoft.kusto.spark.utils.KustoConstants
 
-import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.{Await, Future}
+import scala.concurrent.duration._
+import java.util.concurrent.TimeoutException
 
 private[kusto] case class KustoRelation(kustoCoordinates: KustoCoordinates,
                                         authentication: KustoAuthentication,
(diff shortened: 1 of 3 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
```diff
@@ -30,6 +34,7 @@ private[kusto] case class KustoRelation(kustoCoordinates: KustoCoordinates,
 
   private val normalizedQuery = KustoQueryUtils.normalizeQuery(query)
   var cachedSchema: StructType = _
+  val timeoutForCountCheck : FiniteDuration = 10 seconds
```
danield137: where is this being used?

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
```diff
@@ -45,7 +50,15 @@ private[kusto] case class KustoRelation(kustoCoordinates: KustoCoordinates,
 
   override def buildScan(): RDD[Row] = {
     val kustoClient = KustoClientCache.getClient(kustoCoordinates.cluster, authentication).engineClient
-    val count = KDSU.countRows(kustoClient, query, kustoCoordinates.database)
+
+    var count = 0
+    try {
+      count = Await.result(Future(KDSU.countRows(kustoClient, query, kustoCoordinates.database)), timeout)
```
danield137: where is `timeout` set?

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
```diff
@@ -45,7 +50,15 @@ private[kusto] case class KustoRelation(kustoCoordinates: KustoCoordinates,
...
+    } catch {
+      // Count must be high if took more than 10 seconds
+      case TimeoutException => count = KustoConstants.directQueryUpperBoundRows + 1
```
danield137: this is really confusing. If what you want is to prevent lean mode from running, please use a boolean inseam, and should not be aware of internal logic of `isLeanReadMode` (essentially you are hacking code you shouldn't be aware of here).

I would lift the `useLeanMode` flag above set to true by default, and only change it to false if exception or count is higher than threshold.

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
ohadbitt: my miss - ty

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
ohadbitt: timeout is set by user - but i replaced the usage here with timeoutForCountCheck ..

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
ohadbitt: you are right, check next iteration

PR #80, opened 2019-08-05
write-straight

ohadbitt:
Write straight to buffer writer instead of using strings concatenations
To follow the size of the data allready buffered i use - BufferWriter class with a small additional feature


connector/src/main/java/com/microsoft/kusto/spark/javaUtil/BufferWriterWithProgress.java
```diff
@@ -0,0 +1,289 @@
+package com.microsoft.kusto.spark.javaUtil;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.io.*;
+
+/*
+ * Copyright (c) 1996, 2013, Oracle and/or its affiliates. All rights reserved.
```
danield137: where did you copy this from? what are the license terms?

ohadbitt: its from java.io.BufferWriter class - just added the progress thing

danield137: link?

ohadbitt: to the code itself ? took it from intellij .. but its just this class : https://docs.oracle.com/javase/7/docs/api/java/io/BufferedWriter.html
how am i suppose to do it right ? write it myself would be better ?

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoCsvWriter.scala
```diff
@@ -0,0 +1,52 @@
+package com.microsoft.kusto.spark.datasink
+
+import java.io.Writer
+
+case class KustoCsvWriter(out: Writer) {
```
danield137: I would avoid prefixing with Kusto, the code here has no Kusto specific logic

More like , `CountingCsvWriter`

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoCsvWriter.scala
```diff
@@ -0,0 +1,52 @@
...
+  val newLineSep: String = java.security.AccessController.doPrivileged(
+    new sun.security.action.GetPropertyAction("line.separator"))
+  val newLineSepLength: Int = newLineSep.length
+  var progress: Long = 0L
```
danield137: progress usually is in terms of 0 to 100, or in other words, progress implies it's number is meaningful for progression tracking.. here you are just counting bytes..so I would rename to `bytsCounter` or something similar

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -218,130 +212,140 @@ object KustoWriter {
...
+    val (currentBlobWriter, blobsToIngest) = rows.foldLeft[(BlobWriteResource, Seq[(CloudBlockBlob, Long)])]((initialBlobWriter, Seq())) {
+      case ((blobWriter, blobsCreated), row) =>
+
+        writeRowAsCSV(row, schema, dateFormat, blobWriter.csvWriter)
 
         // Check if we crossed the threshold for a single blob to ingest.
         // If so, add current blob to the list of blobs to ingest, and open a new blob writer
-        if (newTotalSize < maxBlobSize) {
-          (newTotalSize, blobWriter, blobsCreated)
+        if (blobWriter.csvWriter.getProgress < maxBlobSize) {
```
danield137: this helps make the point I wrote above, a more readable version of this would be:
```scala
val shouldRoll = lobWriter.csvWriter.currentSize < maxBlobSize;
if (shouldRoll ) {
```

That way no comment is needed..

connector/src/main/scala/com/microsoft/kusto/spark/datasink/CountingCsvWriter.scala
```diff
@@ -40,13 +40,13 @@ case class KustoCsvWriter(out: Writer) {
       out.write("\"")
     }
 
-    progress += str.length
+    bytsCounter += str.length
   }
 
-  def getProgress: Long = progress
+  def getProgress: Long = bytsCounter
```
danield137: getProgress missed this

PR #89, opened 2019-09-26
Fix date formatting

aelij:


connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -267,14 +267,17 @@ object KustoWriter {
   // This method does not check for null at the current row idx and should be checked before !
   private def writeField(row: SpecializedGetters, fieldIndexInRow: Int, dataType: DataType, dateFormat: FastDateFormat, csvWriter: CountingCsvWriter, nested: Boolean): Unit = {
     dataType match {
-      case DateType => csvWriter.write(DateTimeUtils.toJavaDate(row.getInt(fieldIndexInRow)).toString)
```
ohadbitt: why ? 
writeStringField as stated is only for string fields - as these need to be surrounded by "" and each " needs to be duplicated

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -267,14 +267,17 @@ object KustoWriter {
...
       case StringType => GetStringFromUTF8(row.getUTF8String(fieldIndexInRow), nested, csvWriter)
       case BooleanType => csvWriter.write(row.getBoolean(fieldIndexInRow).toString)
       case structType: StructType => convertStructToCsv(row.getStruct(fieldIndexInRow, structType.length), structType, dateFormat, csvWriter, nested)
       case arrType: ArrayType => convertArrayToCsv(row.getArray(fieldIndexInRow), arrType.elementType, dateFormat, csvWriter, nested)
       case mapType: MapType => convertMapToCsv(row.getMap(fieldIndexInRow), mapType, dateFormat, csvWriter, nested)
-      case _ => csvWriter.write(row.get(fieldIndexInRow, dataType).toString)
+      case ByteType | ShortType | IntegerType | LongType | FloatType | DoubleType =>
+        csvWriter.write(row.get(fieldIndexInRow, dataType).toString)
+      case _: DecimalType => csvWriter.write(row.get(fieldIndexInRow, dataType).toString)
+      case _ => csvWriter.writeStringField(row.get(fieldIndexInRow, dataType).toString, nested)
```
ohadbitt: here too

PR #95, opened 2019-11-10
Use mapping and flushImmediately options over staging table if set

ohadbitt:
Flush immediately if blob larger than 100 mb
Set max blob size to 100 MB
Use mapping and flushImmediately options on the staging table if set by the user

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoSinkOptions.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoSinkOptions.scala
@@ -28,8 +28,8 @@ object KustoSinkOptions extends KustoOptions{
   val NONE_RESULT_LIMIT = "none"
 
   // A limit indicating the size in MB of the aggregated data before ingested to Kusto. Note that this is done for each
-  // partition. The ingestion Kusto also aggregates data, default suggested by Kusto is 1GB but here we suggest to cut
-  // it at 100MG to adjust it to spark pulling of data.
+  // partition. Kusto's ingestion also aggregates data, default suggested by Kusto is 1GB but here we suggest to cut
+  // it at 100MB to adjust it to spark pulling of data.
   val KUSTO_CLIENT_BATCHING_LIMIT: String = newOption("clientBatchingLimit")
 
 }
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoSinkOptions.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoSinkOptions.scala
@@ -26,6 +26,12 @@ object KustoSinkOptions extends KustoOptions{
   val KUSTO_SPARK_INGESTION_PROPERTIES_JSON: String = newOption("sparkIngestionPropertiesJson")
 
   val NONE_RESULT_LIMIT = "none"
+
+  // A limit indicating the size in MB of the aggregated data before ingested to Kusto. Note that this is done for each
+  // partition. The ingestion Kusto also aggregates data, default suggested by Kusto is 1GB but here we suggest to cut
+  // it at 100MG to adjust it to spark pulling of data.
+  val KUSTO_CLIENT_BATCHING_LIMIT: String = newOption("clientBatchingLimit")
+
 }
 
 object SinkTableCreationMode extends Enumeration {
@@ -37,4 +43,5 @@ case class WriteOptions(tableCreateOptions: SinkTableCreationMode.SinkTableCreat
                         isAsync: Boolean = false,
                         writeResultLimit: String = KustoSinkOptions.NONE_RESULT_LIMIT,
                         timeZone: String = "UTC", timeout: FiniteDuration,
-                        IngestionProperties: Option[String] = None)
+                        IngestionProperties: Option[String] = None,
+                        batchLimit: Int = 100)
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
@@ -37,14 +37,14 @@ object KustoWriter {
   val TempIngestionTablePrefix = "_tmpTable"
   val delayPeriodBetweenCalls: Int = KCONST.defaultPeriodicSamplePeriod.toMillis.toInt
   val GZIP_BUFFER_SIZE: Int = KCONST.defaultBufferSize
-  val maxBlobSize: Int = 100 * KCONST.oneMega
+  var maxBlobSize: Int = 100 * KCONST.oneMega
 
   private[kusto] def write(batchId: Option[Long],
                            data: DataFrame,
                            tableCoordinates: KustoCoordinates,
                            authentication: KustoAuthentication,
                            writeOptions: WriteOptions): Unit = {
-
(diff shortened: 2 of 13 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoSinkOptions.scala
```diff
@@ -26,6 +26,12 @@ object KustoSinkOptions extends KustoOptions{
   val KUSTO_SPARK_INGESTION_PROPERTIES_JSON: String = newOption("sparkIngestionPropertiesJson")
 
   val NONE_RESULT_LIMIT = "none"
+
+  // A limit indicating the size in MB of the aggregated data before ingested to Kusto. Note that this is done for each
+  // partition. The ingestion Kusto also aggregates data, default suggested by Kusto is 1GB but here we suggest to cut
+  // it at 100MG to adjust it to spark pulling of data.
```
danield137:
```suggestion
  // it at 100MB to adjust it to spark pulling of data.
```

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoSinkOptions.scala
```diff
@@ -26,6 +26,12 @@ object KustoSinkOptions extends KustoOptions{
   val KUSTO_SPARK_INGESTION_PROPERTIES_JSON: String = newOption("sparkIngestionPropertiesJson")
 
   val NONE_RESULT_LIMIT = "none"
+
+  // A limit indicating the size in MB of the aggregated data before ingested to Kusto. Note that this is done for each
+  // partition. The ingestion Kusto also aggregates data, default suggested by Kusto is 1GB but here we suggest to cut
```
danield137:
```suggestion
  // partition. Kusto's ingestion also aggregates data, default suggested by Kusto is 1GB but here we suggest to cut
```

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -25,26 +25,26 @@ import org.apache.spark.sql.catalyst.util.{ArrayData, DateTimeUtils, MapData}
...
-import scala.concurrent.{Await, Future}
+import scala.concurrent.{Await, Future, TimeoutException}
 
 object KustoWriter {
   private val myName = this.getClass.getSimpleName
   val TempIngestionTablePrefix = "_tmpTable"
   val delayPeriodBetweenCalls: Int = KCONST.defaultPeriodicSamplePeriod.toMillis.toInt
   val GZIP_BUFFER_SIZE: Int = KCONST.defaultBufferSize
-  val maxBlobSize: Int = KCONST.oneGiga - KCONST.oneMega
+  var maxBlobSize: Int = 100 * KCONST.oneMega
```
danield137: doesn't really matter now, but I think the common way is  `1_MB` or `ONE_MB`

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -152,7 +151,19 @@ object KustoWriter {
 
     KDSU.logWarn(myName, s"Ingesting using ingest client - partition: ${TaskContext.getPartitionId()}")
 
-    tasks.asScala.foreach(t => Await.result(t, KCONST.defaultIngestionTaskTime))
+    tasks.asScala.foreach(t => try{
```
danield137:
```suggestion
    tasks.asScala.foreach(t => try {
```

connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
```diff
@@ -94,4 +94,12 @@ private[kusto] object CslCommandsGenerator {
   def generateTableAlterRetentionPolicy(tmpTableName: String, period: String, recoverable: Boolean): String = {
     s""".alter table $tmpTableName policy retention '{ "SoftDeletePeriod": "$period", "Recoverability":"${if (recoverable) "Enabled" else "Disabled"}" }' """
   }
+
+  def generateShowTableMappingsCommand(tableName: String, kind: String): String = {
+    s""".show table $tableName ingestion $kind mappings"""
```
danield137: tablename escaping? also, mapping using new syntax maybe? (optional: maybe move the command generator to our java client? versioning might be easier)

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
```diff
@@ -175,6 +175,27 @@ class KustoClient(val clusterAlias: String, val engineKcsb: ConnectionStringBuil
     }
   }
 
+  private[kusto] def setMappingOnStagingTableIfNeeded(ingestionProperties: IngestionProperties, originalTable: String): Unit = {
+    val mapping = ingestionProperties.getIngestionMapping
+    val mappingReferenceName = mapping.getIngestionMappingReference
+    if (StringUtils.isNotBlank(mappingReferenceName)) {
+      val mappingKind = mapping.getIngestionMappingKind.toString
+      val cmd = generateShowTableMappingsCommand(originalTable, mappingKind)
+      val policies = engineClient.execute(ingestionProperties.getDatabaseName, cmd).getValues
```
danield137: policies? not mappings?

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
```diff
@@ -175,6 +175,27 @@ class KustoClient(val clusterAlias: String, val engineKcsb: ConnectionStringBuil
...
+      val it = policies.iterator()
+      var found = false
+      while (it.hasNext && !found){
+        val policy = it.next()
+        if(policy.get(0).equals(mappingReferenceName)){
+          val policyJson = policy.get(2).replace("\"","'")
+          val c = generateCreateTableMappingCommand(ingestionProperties.getTableName, mappingKind, mappingReferenceName, policyJson)
```
danield137: not sure I follow, what does `ingestionProperties.getTableName` return? the temp table name or the target table name?

connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
ohadbitt: i think its not really nice that every command i add ill have to add it to java first, publish the version there - and only then use it here ..

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
ohadbitt: hh :)

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
ohadbitt: the ingestion is to the tmpTableName - therefore ingestionProperties.getTableName returns the tmpTable

danield137: something about the name is inconsistent. 
function name is `setMappingOnStagingTableIfNeeded`, so i would expect the param name to be `stagingIngestionProperites`, and that the calling code will have some variable named accordingly.

PR #99, opened 2019-12-02
escape in json

ohadbitt:


connector/src/main/scala/com/microsoft/kusto/spark/datasink/CountingCsvWriter.scala
```diff
@@ -33,13 +33,40 @@ case class CountingCsvWriter(out: Writer) {
...
+            bytesCounter += 1
+          } else if (c == '\b'){
+            out.write("\\b")
+            bytesCounter += 1
+          } else if ( c == '\f') {
+            out.write("\\f")
+            bytesCounter += 1
+          } else if (c == '\t'){
+            out.write("\\t")
+            bytesCounter += 1
```
danield137: you have \t twice.
maybe have a collection of escapable chars and iterate them?
```java 
if (nested) {
    for (c <- str) {
        for ( escapable <- ['\t', '\n'...] {
            if c == esacapeable {
                     out.write('\' + c);
             }
        }
    }
}
```

connector/src/main/scala/com/microsoft/kusto/spark/datasink/CountingCsvWriter.scala
```diff
@@ -33,13 +33,40 @@ case class CountingCsvWriter(out: Writer) {
 
       bytesCounter += 2
 
-      for (c <- str) {
-        if (c == '"') {
-          out.write("\"\"")
-          bytesCounter += 1
+      // Json does not allow special characters
+      if(nested){
```
danield137: > Json does not allow special chars

I'm not sure I follow the logic. this function is called writeStringField, and is in a class called CsvWriter , 
why would you get a json?

Maybe we should handle this bad state else where?

ohadbitt: Kusto dynamic is json

connector/src/main/scala/com/microsoft/kusto/spark/datasink/CountingCsvWriter.scala
ohadbitt: This wont work as theres a differnce between
'\r' and '\\r'
One is a special char , the later is slash followed by the letter r

ohadbitt: This wont work as theres a differnce between
'\r' and '\\r'
One is a special char , the later is slash followed by the letter r

connector/src/main/scala/com/microsoft/kusto/spark/datasink/CountingCsvWriter.scala
vladikbr: This is the writeStringField method in class CountingCsvWriter class.
This has absolutely nothing to do with Json.

If it needs to enclose the field in quotes, it should.
If it needs to double the escaping sequences, it should.

You can tale a look at how the C# CsvWriter works and handles JSON string fields for reference.

ohadbitt: I did 
This more resambles our JsonEscapedCsvWriter besides my addition of \f and \b
Do you suggest that the user will have to deal with it?

PR #102, opened 2020-01-19
Export containers

ohadbitt:
By default an estimation of the rows count is first being made, if the count is lower than 5000 records a simple query is made, else - if storage params were provided they are used for scale reading and if not - the connector tries to use storage from the kusto ingest service. 
A new option - `KUSTO_READ_MODE` allows to override the default behavior and force a read mode. 
In default mode if a lean mode was chosen and failed - there is a fallback to scale mode.


connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
```diff
@@ -56,11 +51,6 @@ private[kusto] object KustoDebugOptions {
     name
   }
 
-  // Reading method is determined internally by the connector
-  // This option allows to override connector heuristics and force a specific mode.
-  // Recommended to use only for debug and testing purposes
-  // Supported values: Empty string (""), 'lean' (direct query), 'scale' (via blob). Default: empty
-  val KUSTO_DBG_FORCE_READ_MODE: String = newOption("dbgForceReadMode")
```
ohadbitt: created READ_MODE instead

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoSourceOptions.scala
```diff
@@ -23,4 +24,19 @@ object KustoSourceOptions extends KustoOptions {
...
+  // This option allows to override these connector heuristics.
+  // In default mode if a lean mode was chosen and failed - there is a fallback to scale mode
+  // See https://docs.microsoft.com/en-us/azure/kusto/concepts/querylimits#limit-on-result-set-size-result-truncation
+  // for hard limit on query size using lean mode
+  val KUSTO_READ_MODE: String = newOption("readMode")
 }
+
+object ReadMode extends Enumeration {
+  type ReadMode = Value
+  val Default, ForceLeanMode, ForceScaleMode = Value
```
danield137: Default is not a good value for an Enum, because I have no way of understanding what default is in the context of `ReadMode` ...

connector/src/main/scala/com/microsoft/kusto/spark/utils/ContainersCache.scala
```diff
@@ -0,0 +1,42 @@
...
+
+import com.microsoft.azure.kusto.data.Client
+import com.microsoft.kusto.spark.utils.CslCommandsGenerator.generateCreateTmpStorageCommand
+import com.microsoft.kusto.spark.utils.{KustoDataSourceUtils => KDSU}
+
+import org.joda.time.{DateTime, DateTimeZone, Period}
+
+import scala.collection.JavaConverters._
+
+class ContainersCache[A](val dmClient: Client, val clusterAlias: String, val command: String, cacheEntryCreator: ContainerAndSas => A) {
```
danield137: `ContainerProvider` ?

connector/src/main/scala/com/microsoft/kusto/spark/utils/ContainersCache.scala
```diff
@@ -0,0 +1,42 @@
...
+        KDSU.reportExceptionAndThrow(myName, new RuntimeException("Failed to allocate temporary storage"), "writing to Kusto", clusterAlias)
+      }
+
+      lastRefresh = new DateTime(DateTimeZone.UTC)
+      storageUris = scala.util.Random.shuffle(storage)
+      roundRobinIdx = 0
+      storage(roundRobinIdx)
+    }
+    else {
+      roundRobinIdx = (roundRobinIdx + 1) % storageUris.size
```
danield137: can `storageUris.size` be zero?

connector/src/main/scala/com/microsoft/kusto/spark/utils/ContainersCache.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  def getContainer = {
+    // Refresh if storageExpiryMinutes have passed since last refresh for this cluster as SAS should be valid for at least 120 minutes
+    if (storageUris.isEmpty ||
+      new Period(new DateTime(DateTimeZone.UTC), lastRefresh).getMinutes > KustoConstants.storageExpiryMinutes) {
+
+      val res = dmClient.execute(command)
+      val storage = res.getValues.asScala.map(row => {
+        val parts = row.get(0).split('?')
+        cacheEntryCreator(ContainerAndSas(parts(0), '?' + parts(1)))
```
danield137: `cacheEntryCreator` -> `factoryMethod`?

danield137: plus, why is this a paramater? if it is a ContainerCache, shouldn't it know how to create an entry?

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
```diff
@@ -67,33 +71,14 @@ class KustoClient(val clusterAlias: String, val engineKcsb: ConnectionStringBuil
...
 
-      val res = dmClient.execute(generateCreateTmpStorageCommand())
-      val storage = res.getValues.asScala.map(row => {
-        val parts = row.get(0).split('?'); ContainerAndSas(parts(0), '?' + parts(1))
-      })
 
-      if (storage.isEmpty) {
-        KDSU.reportExceptionAndThrow(myName, new RuntimeException("Failed to allocate temporary storage"), "writing to Kusto", clusterAlias)
-      }
+  def getTempBlobInfoForIngestion: ContainerAndSas = {
```
danield137: why is it called blob info?

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
```diff
@@ -31,6 +31,10 @@ class KustoClient(val clusterAlias: String, val engineKcsb: ConnectionStringBuil
   lazy val dmClient: Client = ClientFactory.createClient(ingestKcsb)
   lazy val ingestClient: IngestClient = IngestClientFactory.createClient(ingestKcsb)
 
+  private lazy val exportCacheEntryCreator = (c: ContainerAndSas) => KDSU.parseSas(c.containerUrl + c.sas)
+  private lazy val  ingestContainersCache = new ContainersCache(dmClient, clusterAlias, generateCreateTmpStorageCommand(), (c: ContainerAndSas) => c)
```
danield137: this might be a silly question, but why do we need to split the ingestion containers from the export ones?

danield137: nvm, noticed the different command

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoSourceOptions.scala
ohadbitt: I dont think i can pass the logic by the name here - the huristics is described in the options ... i will add it to the docs

connector/src/main/scala/com/microsoft/kusto/spark/utils/ContainersCache.scala
ohadbitt: its a cache ... no ?

connector/src/main/scala/com/microsoft/kusto/spark/utils/ContainersCache.scala
ohadbitt: it actually can only if we failed to retrieve the storage .. or if the user doesn't have permissions

connector/src/main/scala/com/microsoft/kusto/spark/utils/ContainersCache.scala
ohadbitt: its generic .. instead i could do like the dm - create a class that would extend the cache and create the entry

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
ohadbitt: its just a url and sas -> should i just drop the `info`?

connector/src/main/scala/com/microsoft/kusto/spark/utils/ContainersCache.scala
danield137: do we handle it properly or will it raise a math exception?

connector/src/main/scala/com/microsoft/kusto/spark/utils/ContainersCache.scala
danield137: if it's generic, then the class can't be called ContainersCache...

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
danield137: yes

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoSourceOptions.scala
danield137: how about - `BestEffort`, or `Optimized`? (I don't like the term Default because it is a reserved keyword in a lot of languages)

connector/src/main/scala/com/microsoft/kusto/spark/utils/ContainersCache.scala
danield137: Usually, a cache doesn't use round-robin to provide you with values... There is some business logic here. so cache seems like it's not providing enough information

connector/src/main/scala/com/microsoft/kusto/spark/utils/ContainersCache.scala
ohadbitt: it is only for containers - one time export containers one time ingestion containers

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoSourceOptions.scala
ohadbitt: i really prefer the default actually - also discussed this with boaz .. your suggestions suggest that there is a third option . i really prefer it being "Default", "Regular" or maybe just drop it and instead check if the option was provided ..

connector/src/main/scala/com/microsoft/kusto/spark/utils/ContainersCache.scala
ohadbitt: we throw .. look at the refresh function

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoSourceOptions.scala
danield137: Ok, so check if it was provided. 
Default is not a real enum value.

connector/src/main/scala/com/microsoft/kusto/spark/utils/ContainersCache.scala
```diff
@@ -0,0 +1,42 @@
...
+      })
+
+      if (storage.isEmpty) {
+        KDSU.reportExceptionAndThrow(myName, new RuntimeException("Failed to allocate temporary storage"), "writing to Kusto", clusterAlias)
+      }
+
+      lastRefresh = new DateTime(DateTimeZone.UTC)
+      storageUris = scala.util.Random.shuffle(storage)
```
danield137: you don't need to shuffle the array, you can just pick a random number, no?

PR #104, opened 2020-01-30
Hide table

ohadbitt:
Create temporary tables as hidden

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
@@ -55,7 +55,7 @@ class KustoClient(val clusterAlias: String, val engineKcsb: ConnectionStringBuil
           tableSchemaBuilder.add(s"['${field.name}']:$fieldType")
         }
         tmpTableSchema = tableSchemaBuilder.toString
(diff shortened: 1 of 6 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
```diff
@@ -55,15 +55,15 @@ class KustoClient(val clusterAlias: String, val engineKcsb: ConnectionStringBuil
...
+        engineClient.execute(database, generateTableCreateCommand(table, tmpTableSchema, hidden = false))
       }
     } else {
       // Table exists. Parse kusto table schema and check if it matches the dataframes schema
       tmpTableSchema = extractSchemaFromResultTable(schemaShowCommandResult)
     }
 
     //  Create a temporary table with the kusto or dataframe parsed schema with 1 day retention
-    engineClient.execute(database, generateTableCreateCommand(tmpTableName, tmpTableSchema))
+    engineClient.execute(database, generateTableCreateCommand(tmpTableName, tmpTableSchema, hidden = true))
```
danield137: I would suggest instead of a flag, use a different command 
for example `generateTempTableCreateCommand`, which has the property of being hidden, but is more readable IMHO

PR #106, opened 2020-02-18
Docs

ohadbitt:
Just for sdk and docs

### DIFF
--- a/docs/KustoSource.md
+++ b/docs/KustoSource.md
@@ -21,7 +21,7 @@ that is using it. Please verify the following before using Kusto connector:
  
  Kusto connector implements Spark 'Datasource V1' API. 
  Kusto data source identifier is "com.microsoft.kusto.spark.datasource". 
- Kusto table schema is translated into a spark as explained in [Spark-Kusto DataTypes](mapping.md).
+ Kusto table schema is translated into a spark schema as explained in [Spark-Kusto DataTypes](mapping.md).
  The reading is done in one of two modes: 'Single' and 'Distributed'. By default if the user did not override the read mode - an estimated
  count for the query is made; for small row count - a 'Single' mode is chosen with a fallback to the 'Distributed' mode (as the first might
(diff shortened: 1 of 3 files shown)


docs/KustoSource.md
```diff
@@ -21,7 +21,7 @@ that is using it. Please verify the following before using Kusto connector:
  
  Kusto connector implements Spark 'Datasource V1' API. 
  Kusto data source identifier is "com.microsoft.kusto.spark.datasource". 
- Kusto table schema is translated into a spark as explained in [DataTypes](Spark-Kusto DataTypes mapping.md).
+ Kusto table schema is translated into a spark as explained in [Spark-Kusto DataTypes](mapping.md).
```
danield137: >  into a spark 
what does this mean?

docs/KustoSource.md
```diff
@@ -35,15 +35,16 @@ that is using it. Please verify the following before using Kusto connector:
...
   ```
  where:
  * **Kusto-query** is any valid Kusto query. For details, please refer to [Query statements documentation](https://docs.microsoft.com/en-us/azure/kusto/query/statements). To read the whole table, just provide the table name as query.
  >Note:
  Kusto commands  cannot be executed via the connector.
  * **Parameters map** is a set of key-value pairs, where the key is the parameter name. See [Supported Options](#supported-options)
  section for details
- 
+ * **ClientRequestProperties** is a json of [ClientRequestProperties](https://github.com/Azure/azure-kusto-java/blob/master/data/src/main/java/com/microsoft/azure/kusto/data/ClientRequestProperties.java) object - 
+ optional. Parameters described [here](https://docs.microsoft.com/en-us/azure/kusto/api/netfx/request-properties).
```
danield137: better to start with saying it's optional, then a short desc, then a link

docs/KustoSource.md
ohadbitt: nothing

danield137: hmm was it supposed to say `into a spark scheme`?

docs/KustoSource.md
ohadbitt: this is actually a copy paste mistake

PR #108, opened 2020-02-25
Timespan as to string

ohadbitt:
Timespan did not work well in distributed mode. In a simple query timespan returns from kusto as a string, but in export it is exported as long (Ticks..)


connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoReader.scala
```diff
@@ -44,29 +44,29 @@ private[kusto] case class KustoReadRequest(sparkSession: SparkSession,
...
   }
 
   private[kusto] def distributedBuildScan(kustoClient: Client,
                                           request: KustoReadRequest,
                                           storage: Seq[KustoStorageParameters],
                                           partitionInfo: KustoPartitionParameters,
                                           options: KustoReadOptions,
-                                          filtering: KustoFiltering): RDD[Row] = {
+                                          filtering: KustoFiltering,
+                                          timespanColumns: Set[String]): RDD[Row] = {
```
danield137: here as well, kustoScheme? it seems too specific to pass this

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoFilter.scala
```diff
@@ -7,18 +7,18 @@ import org.apache.spark.sql.types._
...
 
   def pruneSchema(schema: StructType, columns: Array[String]): StructType = {
     val fieldMap = schema.fields.map(x => x.name -> x).toMap
     new StructType(columns.map(name => fieldMap(name)))
   }
 
-  def buildColumnsClause(columns: Array[String]): String = {
-    if(columns.isEmpty) "" else {
-      " | project " + columns.map(col=>s"['$col']").mkString(", ")
+  def buildColumnsClause(columns: Array[String], timespanColumns: Set[String]): String = {
```
danield137: name here should be more aligned with what is happening, something like `toStringColumns`.

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoReader.scala
```diff
@@ -179,12 +179,13 @@ private[kusto] class KustoReader(client: Client, request: KustoReadRequest, stor
                                            directory: String,
                                            options: KustoReadOptions,
                                            filtering: KustoFiltering,
-                                           clientRequestProperties: Option[ClientRequestProperties]): Unit = {
+                                           clientRequestProperties: Option[ClientRequestProperties],
+                                           timespanColumns: Set[String]): Unit = {
```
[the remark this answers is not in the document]
ohadbitt: i don't see why 
schema (which is part of kustoSchema) is part of the KustoReadRequest  - which is passed here already ...

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoReader.scala
ohadbitt: answered there

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoFilter.scala
ohadbitt: called it toStringCastedColumns

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoReader.scala
danield137: mostly because it is strange that a function called `distributedBuildScan` will require knowledge of timespanColumns. it will make more sense down the line

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoReader.scala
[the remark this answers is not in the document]
ohadbitt: changed it so kustoSchema  is part of the request

PR #111, opened 2020-03-22
Normalize database name

ohadbitt:
Database with whitespaces failed for blob name includes it

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
@@ -183,7 +183,7 @@ object KustoWriter {
                        tableCoordinates: KustoCoordinates,
                        tmpTableName: String,
                        client: KustoClient): BlobWriteResource = {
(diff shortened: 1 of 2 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -183,7 +183,7 @@ object KustoWriter {
                        tableCoordinates: KustoCoordinates,
                        tmpTableName: String,
                        client: KustoClient): BlobWriteResource = {
-    val blobName = s"${tableCoordinates.database}_${tmpTableName}_${UUID.randomUUID.toString}_spark.csv.gz"
+    val blobName = s"${KustoQueryUtils.simplifyName(tableCoordinates.database)}_${tmpTableName}_${UUID.randomUUID.toString}_spark.csv.gz"
```
vladikbr: This should be normalizeTableName, shouldn't it?
Also, if that function relates to any Kusto entity name, it should be named properly

ohadbitt: normalizeTableName doesn't remove whitesspaces and a blob name cannot have whitespaces

PR #123, opened 2020-05-04
Include Spark 2.4.5 as well

Fokko:
We're using Spark 2.4.5


ohadbitt:
sorry it took us so much time to see
anyway i dont see why this should bother you in your work .. as all the libraries using spark.version property are under 'provided' scope so if you add your version in your dependencies it will take yours

PR #132, opened 2020-06-08
Writing special chars 

ohadbitt:
#### Pull Request Description

Writing chars with value bigger than 128 inside json objects was raising arrayOutOfBounds as we check if it should be escaped
---

#### Future Release Comment
**Fixes:**
- Writing chars with value bigger than 128 inside json objects was raising arrayOutOfBounds as we check if it should be escaped

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/Writers.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/Writers.scala
@@ -62,7 +62,7 @@ case class EscapedWriter(out: java.io.Writer) extends Writer {
 
   def write(str: String): Unit ={
     for (c <- str) {
-      val escaped = EscapedWriter.escapeTable(c)
+      val escaped =  if (c > 127) 0 else EscapedWriter.escapeTable(c)
       if (escaped != 0) {
         out.write('\\')
         out.write(escaped)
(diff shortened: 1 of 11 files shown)


samples/src/main/scala/KustoConnectorDemo.scala
```diff
@@ -58,7 +58,7 @@ object KustoConnectorDemo {
       .option(KustoSinkOptions.KUSTO_CLUSTER, cluster)
       .option(KustoSinkOptions.KUSTO_DATABASE, database)
       .option(KustoSinkOptions.KUSTO_TABLE, table)
-      .option(KustoSinkOptions.KUSTO_AAD_CLIENT_ID, appId)
+      .option(KustoOptions.KUSTO_AAD_CLIENT_ID, appId)
```
vladikbr, line 61: Isn't this supposed to be KustoSinkOptions?

connector/src/main/scala/com/microsoft/kusto/spark/datasink/Writers.scala
```diff
@@ -62,7 +62,7 @@ case class EscapedWriter(out: java.io.Writer) extends Writer {
 
   def write(str: String): Unit ={
     for (c <- str) {
-      val escaped = EscapedWriter.escapeTable(c)
+      val escaped =  if (c > 127) 0 else EscapedWriter.escapeTable(c)
```
vladikbr, line 65: So we are dumping all characters over 127? I thought we support UTF8

ohadbitt: every character that needs escaping is less than 127  ..

PR #141, opened 2020-07-29
decimal writing

ohadbitt:
#### Pull Request Description

Write Decimal type as string without scientific notation - with better performance - taken from pq2json

---

#### Future Release Comment

**Breaking Changes:**
- None

**Features:**
- Decimal type - written as string

**Fixes:**
- None

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
@@ -1,6 +1,7 @@
 package com.microsoft.kusto.spark.datasink
 
 import java.io._
+import java.math.BigInteger
 import java.net.URI
 import java.nio.charset.StandardCharsets
 import java.security.InvalidParameterException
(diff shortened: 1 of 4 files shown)


connector/src/test/scala/com/microsoft/kusto/spark/WriterTests.scala
```diff
@@ -152,6 +153,17 @@ class WriterTests extends FlatSpec with Matchers {
       Row(Row(null,""))
     )
 
+    val someData4 = List(
```
nir-boger, line 156: I would suggest to rename it to: someDecimalData

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -397,6 +398,42 @@ object KustoWriter {
     writer.out.toString
   }
 
+  private def writeDecimalField(row: SpecializedGetters, fieldIndexInRow: Int, precision: Int,scale: Int, writer: Writer) = {
```
nir-boger, line 401: What about moving this method (and the other write* method to another class? (Or it can be like an extension method in C#). Will make this file smaller

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -397,6 +398,42 @@ object KustoWriter {
...
+    writer.write('"')
```
nir-boger, line 402: Although you are highly covered with E2E tests I would think it will be better to test the decimal to string representation using a simple unit tests. LMK WDYT

ohadbitt: The tests under WriterTests are not E2E . I added there a test in this PR already

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
ohadbitt: Tried with extension methods over SpecializedGetters but it really made it complex ,
ill take your other suggestion and put in another class

PR #142, opened 2020-08-06
Support non-public domains

ohadbitt:
#### Pull Request Description
Support non-public domains - writing and reading
---

#### Future Release Comment

**Features:**
- Added new option 
-  KUSTO_BLOB_STORAGE_ENDPOINT_SUFFIX "blobStorageEndpointSuffix" - when providing the storage yourself and using storage key for storage parameters, user can provide this option so that the created config will use the right domain

**Fixes:**
- Non-public domains were not parsed right
- Extract domain from blob SAS so that it is set right to the spark config

This goes along with our current Java support for authentication providing login url of the cloud in the environment variable : AadAuthorityUri

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
@@ -35,7 +35,7 @@ trait KustoOptions {
   val KUSTO_TIMEOUT_LIMIT: String = newOption("timeoutLimit")
 }
 
-case class KustoCoordinates(cluster: String, database: String, table: Option[String] = None)
+case class KustoCoordinates(clusterUrl: String, clusterAlias:String, database: String, table: Option[String] = None)
 
 /** **********************************************************************************/
 /*                                    NOTE!!!                                       */
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
@@ -79,7 +79,7 @@ object KustoWriter {
       asyncWork.onFailure {
         case exception: Exception =>
           kustoClient.cleanupIngestionByproducts(tableCoordinates.database, kustoClient.engineClient, tmpTableName)
-          KDSU.reportExceptionAndThrow(myName, exception, "writing data", tableCoordinates.clusterAlias, tableCoordinates.database, table, shouldNotThrow = true)
+          KDSU.reportExceptionAndThrow(myName, exception, "writing data", tableCoordinates.clusterUrl, tableCoordinates.database, table, shouldNotThrow = true)
           KDSU.logError(myName, "The exception is not visible in the driver since we're in async mode")
       }
     } else {
@@ -124,7 +124,7 @@ object KustoWriter {
         KDSU.reportExceptionAndThrow(
           myName,
           ex,
-          "trying to drop temporary tables", coordinates.clusterAlias, coordinates.database, coordinates.table.getOrElse("Unspecified table name"),
+          "trying to drop temporary tables", coordinates.clusterUrl, coordinates.database, coordinates.table.getOrElse("Unspecified table name"),
           shouldNotThrow = true
         )
     }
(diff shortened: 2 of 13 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
```diff
@@ -35,7 +35,7 @@ trait KustoOptions {
   val KUSTO_TIMEOUT_LIMIT: String = newOption("timeoutLimit")
 }
 
-case class KustoCoordinates(cluster: String, database: String, table: Option[String] = None)
+case class KustoCoordinates(clusterUrl: String, clusterAlias:String, database: String, table: Option[String] = None)
```
vladikbr, line 38: Why do we need both clusterAlias and clustyerUrl?

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -79,7 +79,7 @@ object KustoWriter {
       asyncWork.onFailure {
         case exception: Exception =>
           kustoClient.cleanupIngestionByproducts(tableCoordinates.database, kustoClient.engineClient, tmpTableName)
-          KDSU.reportExceptionAndThrow(myName, exception, "writing data", tableCoordinates.cluster, tableCoordinates.database, table, shouldNotThrow = true)
+          KDSU.reportExceptionAndThrow(myName, exception, "writing data", tableCoordinates.clusterAlias, tableCoordinates.database, table, shouldNotThrow = true)
```
vladikbr, line 82: I think we need the full cluster Url here

vladikbr: and below - errors need to contain full cluster URL

connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
ohadbitt: We use alias for logs 
Before we were just keeping the alias but you cant construct the url from just the alias

ohadbitt: Alias is used also as key for the cached client

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoDataSourceUtils.scala
```diff
@@ -28,9 +28,13 @@ import org.apache.commons.lang3.StringUtils
 
 object KustoDataSourceUtils {
   private val klog = Logger.getLogger("KustoConnector")
-  val urlPattern: Regex = raw"https://(?:ingest-)?(.+).kusto.windows.net(?::443)?".r
-  val sasPattern: Regex = raw"(?:https://)?([^.]+).blob.core.windows.net/([^?]+)?(.+)".r
 
+  val sasPattern: Regex = raw"(?:https://)?([^.]+).blob.([^/]+)/([^?]+)?(.+)".r
+  val ingestPrefix = "https://ingest-"
```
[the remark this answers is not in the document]
ohadbitt: mgmt is the alias for ingest clusters

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoDataSourceUtils.scala
```diff
@@ -28,9 +28,13 @@ import org.apache.commons.lang3.StringUtils
...
+  val enginePrefix = "https://"
+  val defaultDomainPostfix = "core.windows.net"
```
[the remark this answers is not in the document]
ohadbitt: sure

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoDataSourceUtils.scala
```diff
@@ -226,17 +231,28 @@ object KustoDataSourceUtils {
...
-      case _ => s"https://$alias.kusto.windows.net"
+  private[kusto] def getEngineUrlFromAliasIfNeeded(cluster: String): String = {
+    if(cluster.startsWith(enginePrefix)){
+      if(cluster.startsWith(ingestPrefix)){
+        cluster.replace(ingestPrefix, "https://")
+      } else{
+        cluster
+      }
+    } else {
+      s"https://$cluster.kusto.windows.net"
```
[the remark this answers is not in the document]
ohadbitt: i have only for the https but ill create another one

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoDataSourceUtils.scala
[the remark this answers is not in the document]
ohadbitt: added to app.properties file we use for versioning

PR #152, opened 2020-09-21
spark 3 scala 2.12 support

ohadbitt:
#### Pull Request Description

spark 3 scala 2.12 support

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ In order to use the connector, you need to have:
 
 - Java 1.8 SDK installed
 - [Maven 3.x](https://maven.apache.org/download.cgi) installed
-- Spark version 2.4.x 
+- Spark version 2.4.x. For spark 3.x use connector versions x.x.x-spark_3.0_2.12.
 
 > Note: when working with 2.3 Spark version or lower, please refer to [Building for legacy Spark versions](docs/CHANGELIST.md#building-for-legacy-spark-versions)
  section of the [CHANGELIST](docs/CHANGELIST.md) document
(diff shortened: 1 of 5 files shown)


ohadbitt:
Open discussion: 
should we actually change our artifact id to reflect the scala version instead of the library version ? 
other libraries have a different artifactId per scala version, but we didnt use it like that from the begining .. should we release from now on two artifacts? as in: 
  spark-kusto-connector_2.11 and spark-kusto-connector_2.12
or should we leave the old one untouched ?

vladikbr:
> Open discussion:
> should we actually change our artifact id to reflect the scala version instead of the library version ?
> other libraries have a different artifactId per scala version, but we didnt use it like that from the begining .. should we release from now on two artifacts? as in:
> spark-kusto-connector_2.11 and spark-kusto-connector_2.12
> or should we leave the old one untouched (currently used in this PR)?

I don't know. Who knows what is the common practice?

anagha-microsoft:
> Open discussion:
> should we actually change our artifact id to reflect the scala version instead of the library version ?
> other libraries have a different artifactId per scala version, but we didnt use it like that from the begining .. should we release from now on two artifacts? as in:
> spark-kusto-connector_2.11 and spark-kusto-connector_2.12
> or should we leave the old one untouched (currently used in this PR)?

@ohadbitt -
1.  Per our discussion, this would be self-evident and user-friendly.
kusto-spark_<n.n.n for spark version>_<n.n for Scala version only>
E.g. kusto-spark_3.0.0_2.12

2.  And we can do this going forward

3.  And firm up the docs to emphasize this change

anagha-microsoft (approved):
This construct is great-
kusto-spark_${spark.version.major}_${scala.version.major}

PR #153, opened 2020-10-04
Create SynapseSample.py

ManojRaheja:
Added Sample file for Synapse

#### Pull Request Description

Added Sample code for Synapse

---

#### Future Release Comment
Added sample pyspark code to demonstrate use of link service and integrated Kusto Spark connector in Azure Synapse. 

**Breaking Changes:**
- None

**Features:**
- Azure Data Explorer connector for Azure Synapse

**Fixes:**
- None

### DIFF
--- a/samples/src/main/python/SynapseSample.py
+++ b/samples/src/main/python/SynapseSample.py
@@ -1,6 +1,9 @@
 # Optional:
-sc._jvm.com.microsoft.kusto.spark.utils.KustoDataSourceUtils.setLoggingLevel("all")
+sc._jvm.com.microsoft.kusto.spark.utils.KustoDataSourceUtils.setLoggingLevel("all") 
 
+# Azure Synapse is a limitless analytics service that brings together enterprise data warehousing. 
+# The synapse-kusto connector encapsulates the kusto-spark connector. The code resides in a private repository and is integrated in azure synapse.
+# Read more about this integration https://docs.microsoft.com/azure/synapse-analytics/quickstart-connect-azure-data-explorer
 
 #######################
 # BATCH READ EXAMPLE  #
@@ -47,7 +50,7 @@ display(kustoDf)
 # BATCH WRITE EXAMPLE #
 #######################
 
-# Write data to a Azure Data Explorer table
+# Write data to an Azure Data Explorer table
 
 df.write \
     .format("com.microsoft.kusto.spark.synapse.datasource") \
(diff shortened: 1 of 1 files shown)


samples/src/main/python/SynapseSample.py
```diff
@@ -0,0 +1,122 @@
+# Optional:
+sc._jvm.com.microsoft.kusto.spark.utils.KustoDataSourceUtils.setLoggingLevel("all")
+
+
+#######################
+# BATCH READ EXAMPLE  #
+#######################
+
+# Read data from Azure Data Explorer table(s)
+# In this example the Kusto Spark connector will determing the optimal path to get data: API for small data sets, Export/Distributed mode for large datasets.
```
ohadbitt, line 10:
```suggestion
# In this example the Kusto Spark connector will determine the optimal path to get data: API for small data sets, Export/Distributed mode for large datasets.
```

samples/src/main/python/SynapseSample.py
```diff
@@ -0,0 +1,122 @@
+# Optional:
```
ohadbitt, line 1: Should we explain that this is not the repository that holds the synapse integration code? we could say that the other package utilizes this one and it is private

samples/src/main/python/SynapseSample.py
```diff
@@ -0,0 +1,122 @@
...
+            .load()
+
+display(kustoDf)    
+
+
+#######################
+# BATCH WRITE EXAMPLE #
+#######################
+
+# Write data to a Azure Data Explorer table
```
ohadbitt, line 50:
```suggestion
# Write data to an Azure Data Explorer table
```

samples/src/main/python/SynapseSample.py
```diff
@@ -0,0 +1,122 @@
...
+
+kustoQ = dfStream \
+    .writeStream \
+    .format("com.microsoft.kusto.spark.synapse.datasink.KustoSynapseSinkProvider") \
+    .option("spark.synapse.linkedService", "<link service name>") \
+    .option("kustoDatabase", "<Database name>") \
+    .option("kustoTable", "<Table name>") \
+    .trigger(once = True)
+    
+kustoQ.start().awaitTermination(60*8)
```
ohadbitt, line 122: i think its better here to not have an await time - as usually when working from databricks/synapse we dont won't to stop the streaming job

ohadbitt (approved):

samples/src/main/python/SynapseSample.py
ManojRaheja: As discussed, please update as you think fit.

PR #158, opened 2020-11-04
Fix cleanup tables + table deletion exception should throw

ohadbitt:
#### Pull Request Description

If the connector couldn't delete the temporary table it used to suppress it.
When the connector starts it has an initial check for old tables - this check did not include hidden tables (which we moved to using at some point) and also it could delete new tables.
---

#### Future Release Comment
**Breaking Changes:**
- Throw on temporary table deletion failure

**Fixes:**
- Initial cleanup of old temporary tables (older than three days)

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
@@ -167,6 +167,7 @@ class KustoClient(val clusterAlias: String, val engineKcsb: ConnectionStringBuil
   private[kusto] def cleanupIngestionByproducts(database: String, kustoAdminClient: Client, tmpTableName: String): Unit = {
     try {
       kustoAdminClient.execute(database, generateTableDropCommand(tmpTableName))
+      KDSU.logInfo(myName, s"Temporary table '$tmpTableName' deleted successfully")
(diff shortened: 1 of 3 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
```diff
@@ -167,10 +167,39 @@ class KustoClient(val clusterAlias: String, val engineKcsb: ConnectionStringBuil
...
+        if (tablesToCleanup.nonEmpty) {
+          engineClient.execute(coordinates.database, generateDropTablesCommand(tablesToCleanup.mkString(",")))
+        }
+      }
+    } onFailure {
+      case ex: Exception =>
+        KDSU.reportExceptionAndThrow(
+          myName,
+          ex,
+          "trying to drop temporary tables", coordinates.clusterUrl, coordinates.database, coordinates.table.getOrElse("Unspecified table name"),
```
vladikbr, line 200: Don't you want to emit an explicit error for each table that we failed to delete?
Like error per table with full details of the table that failed to be deleted.

vladikbr (changes requested):
See comment

vladikbr (approved):

PR #159, opened 2020-11-11
Handle ingestIfNotExists tags

ohadbitt:
IngestIfNotExists tags can't be part of the regular flow as the data is ingested by multiple partitions into a staging table.
The solution here is dealing with the tags before and after the ingestion: First check the table before the ingestion into the staging table so as to skip it at all if the ingestBy tags itersect with the IngestIfNotExists ones. Second - when merging the tables - we do it in a single command that takes into consideration the current tags at this stage.

**Fixes:**
- Handle ingestIfNotExists tags

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
@@ -51,45 +51,57 @@ object KustoWriter {
     val tmpTableName: String = KustoQueryUtils.simplifyName(TempIngestionTablePrefix +
       data.sparkSession.sparkContext.appName +
       "_" + table + batchId.map(b=>s"_${b.toString}").getOrElse("") + "_" + writeOptions.requestId)
+
     implicit val parameters: KustoWriteResource = KustoWriteResource(authentication, tableCoordinates, data.schema, writeOptions, tmpTableName)
-
-    kustoClient.cleanupTempTables(tableCoordinates)
-
-    // KustoWriter will create a temporary table ingesting the data to it.
-    // Only if all executors succeeded the table will be appended to the original destination table.
-    kustoClient.createTmpTableWithSameSchema(tableCoordinates, tmpTableName, writeOptions.tableCreateOptions, data.schema)
-    KDSU.logInfo(myName, s"Successfully created temporary table $tmpTableName, will be deleted after completing the operation")
-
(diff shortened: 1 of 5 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -51,45 +51,57 @@ object KustoWriter {
...
+    // Remove the ingestIfNotExists tags because several partitions can ingest into the staging table and should not interfere one another
+    stagingTableIngestionProperties.setIngestIfNotExists(new util.ArrayList())
+    val shouldIngest = kustoClient.shouldIngestData(tableCoordinates.database, table, writeOptions.IngestionProperties)
 
-    // KustoWriter will create a temporary table ingesting the data to it.
-    // Only if all executors succeeded the table will be appended to the original destination table.
-    kustoClient.createTmpTableWithSameSchema(tableCoordinates, tmpTableName, writeOptions.tableCreateOptions, data.schema)
-    KDSU.logInfo(myName, s"Successfully created temporary table $tmpTableName, will be deleted after completing the operation")
+    if (!shouldIngest) {
+      KDSU.logInfo(myName, s"Ingestion skipped: IngestByTags were given and destination table ingest-by tags intersect with them.")
```
ohadbitt, line 64: "Provided ingest-By tags are present in the destination table $desttable

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
```diff
@@ -141,8 +141,10 @@ class KustoClient(val clusterAlias: String, val engineKcsb: ConnectionStringBuil
         if (partitionsResults.value.size > 0) {
           // Move data to real table
           // Protect tmp table from merge/rebuild and move data to the table requested by customer. This operation is atomic.
+          // We are using the ingestIfNotExists Tags here too (on top of the check at the start of the flow) so that if
+          // several flows started together only one of them would ingest
           kustoAdminClient.execute(database, generateTableAlterMergePolicyCommand(tmpTableName, allowMerge = false, allowRebuild = false))
-          kustoAdminClient.execute(database, generateTableMoveExtentsCommand(tmpTableName, table.get))
+          kustoAdminClient.execute(database, generateTableMoveExtentsCommand(tmpTableName, table.get ,ingestIfNotExistsTags))
```
ohadbitt, line 147: take the results and see if there are any extents there were moved - log it

vladikbr (approved):

PR #165, opened 2020-12-31
DistributedMode - Export request query only once

abmo-x:
#### Pull Request Description

In Distributed mode, requested Kusto query is only exported once if user sets a KUSTO_DISTRIBUTED_MODE_EXPORT_QUERY_ONLY_ONCE=true.

Current behavior
-----------------------
When a dataset is read in distributed mode and not cached as below:

`val df = spark.read.kusto(cluster, database, "MyKustoTable | where (ColB % 1000 == 0) | distinct ColA ", conf)`

Any queries or actions on 'df' will result in an 'export to storage account' and then 'read from storage paths' by spark as parquet rdd.

Feature behavior
---------------
In the same scenario above, when users don't want their large datasets to be persisted and exported multiple times, users can set KUSTO_DISTRIBUTED_MODE_EXPORT_QUERY_ONLY_ONCE=true.

With this flag, data is exported only once from Kusto for a request Query. 
Once exported, all 'df' queries or actions are performed on the exported 'paths' in storage directly via Spark rdd

---

#### Future Release Comment
To export request query data to storage only once in distributed mode,
set KUSTO_DISTRIBUTED_MODE_EXPORT_QUERY_ONLY_ONCE=true 

**Breaking Changes:**
- None

**Features:**
- Export only once per request query in distributed mode.

**Fixes:**
- None


connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoReader.scala
```diff
@@ -69,6 +72,48 @@ private[kusto] object KustoReader {
                                           options: KustoReadOptions,
                                           filtering: KustoFiltering): RDD[Row] = {
     KDSU.logInfo(myName, s"Starting exporting data from Kusto to blob storage. requestId: ${request.requestId}")
+    var paths: Seq[String] = Seq()
+    // if exportQueryOnlyOnce is set to true, then check if path is cached and use it
+    // if not export and cache the path for reuse later
+    if (options.exportQueryOnlyOnce) {
```
ohadbitt, line 78: we should log that we are using the cache and weather or not did we get a cache hit

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoReader.scala
```diff
@@ -44,9 +45,11 @@ private[kusto] case class KustoReadRequest(sparkSession: SparkSession,
 
 private[kusto] case class KustoReadOptions(readMode: Option[ReadMode] = None,
                                            shouldCompressOnExport: Boolean = true,
-                                           exportSplitLimitMb: Long = 1024)
+                                           exportSplitLimitMb: Long = 1024,
+                                           exportQueryOnlyOnce: Boolean = false)
 private[kusto] object KustoReader {
   private val myName = this.getClass.getSimpleName
+  private val cachedQueryPaths = new ConcurrentHashMap[String, Seq[String]]().asScala
```
ohadbitt, line 52: the map should take into account also the KustoCoordinates and authenticaiton - as we do in "KustoRelation.Equals" method

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoSourceOptions.scala
```diff
@@ -35,6 +35,8 @@ object KustoSourceOptions extends KustoOptions {
   // See https://docs.microsoft.com/en-us/azure/kusto/concepts/querylimits#limit-on-result-set-size-result-truncation
   // for hard limit on query size using single mode
   val KUSTO_READ_MODE: String = newOption("readMode")
+  // set to 'true' to export request Query only once and reuse the exported path to read as RDD
+  val KUSTO_DISTRIBUTED_MODE_EXPORT_QUERY_ONLY_ONCE: String = newOption("distributedModeExportQueryOnlyOnce")
```
ohadbitt, line 39: Name chose: KUSTO_DISTRIBUTED_READ_MODE_TRANSIENT_CACHE

ohadbitt (commented):
All in all it looks good - i think we should add another option that could come handy together with this one - which should be similar to this : 
spark.databricks.sqldw.pushdown
To disable our predicate pushdown.
If you choose not to implement this we will do it on next PR .

abmo-x:
> All in all it looks good - i think we should add another option that could come handy together with this one - which should be similar to this :
> spark.databricks.sqldw.pushdown
> To disable our predicate pushdown.
> If you choose not to implement this we will do it on next PR .

updated to address comments.

Regarding "spark.databricks.sqldw.pushdown" will this take affect if its adding on a spark context which is already initialized? 
I haven't tested this, not adding it as part of this PR

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoReader.scala
```diff
@@ -68,6 +75,51 @@ private[kusto] object KustoReader {
                                           partitionInfo: KustoPartitionParameters,
                                           options: KustoReadOptions,
                                           filtering: KustoFiltering): RDD[Row] = {
+    var paths: Seq[String] = Seq()
+    // if distributedReadModeTransientCacheEnabled is set to true, then check if path is cached and use it
+    // if not export and cache the path for reuse
+    if (options.distributedReadModeTransientCacheEnabled) {
+      val key = DistributedReadModeTransientCacheKey(request.query, request.kustoCoordinates, request.authentication)
+      if (distributedReadModeTransientCache.contains(key)) {
+        KDSU.logInfo(myName, s"distributedReadModeTransientCache: ${options.distributedReadModeTransientCacheEnabled} cache: hit, reusing cached export paths")
```
ohadbitt, line 84: options.distributedReadModeTransientCacheEnabled is always true here - same for the other logs .. the value is already know )
```suggestion
        KDSU.logInfo(myName, s"Fetching from distributedReadModeTransientCache: hit, reusing cached export paths")
```

abmo-x: updated, removed the else block log as exportToStorage will log exporting log.

ohadbitt (approved):

PR #179, opened 2021-04-12
Dependencies and some styling

ohadbitt:
**Fixes:**
- Exclude conflicting dependencies

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/utils/DataTypeMapping.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/utils/DataTypeMapping.scala
@@ -5,7 +5,7 @@ import org.apache.spark.sql.types.{ArrayType, DataType, DataTypes, DecimalType, 
 
 object DataTypeMapping {
 
-  val kustoTypeToSparkTypeMap: Map[String, DataType] = Map(
+  val KustoTypeToSparkTypeMap: Map[String, DataType] = Map(
     "string" -> StringType,
     "long" -> LongType,
     "datetime" -> TimestampType,// Kusto datetime is equivalent to TimestampType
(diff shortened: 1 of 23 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/utils/DataTypeMapping.scala
```diff
@@ -32,7 +32,7 @@ object DataTypeMapping {
     "object" -> StringType
   )
 
-  val sparkTypeToKustoTypeMap: Map[DataType, String] = Map(
+  val SparkTypeToKustoTypeMap: Map[DataType, String] = Map(
```
yihezkel, line 35: This seems to duplicate KustoTypeToSparkTypeMap
THis is out of scope for your change, but just a note that it seems we can use Guava's BiMap here to remove the duplication.

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoConstants.scala
```diff
@@ -4,18 +4,18 @@ import scala.concurrent.duration._
...
-  val oneMega: Int = oneKilo * 1024
-  val oneGiga: Int = oneMega * 1024
+  val DefaultWaitingIntervalLongRunning: String = (2 days).toSeconds.toString
+  val DefaultPeriodicSamplePeriod: FiniteDuration = 1 seconds
+  val DefaultIngestionTaskTime: FiniteDuration = 20 seconds
+  val ClientName: String = KustoDataSourceUtils.clientName
+  val DefaultBufferSize: Int = 16 * 1024
+  val StorageExpiryMinutes: Int = 120
+  val SparkSettingsRefreshMinutes: Int = 120
+  val OneKilo: Int = 1024
```
yihezkel, line 14: nit: units are left out. Recommend including the unit (e.g. OneKilobit)

yihezkel (approved):

connector/src/main/scala/com/microsoft/kusto/spark/utils/DataTypeMapping.scala
ohadbitt: its not 1 to 1

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoConstants.scala
ohadbitt: OneKilobite ...

PR #185, opened 2021-05-27
Timestamp accuracy

ohadbitt:
#### Pull Request Description

Timestamp accuracy is not achieved using the current date formatter 
---

**Features:**
- Timestamp accuracy is not achieved using the current date formatter

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/RowCSVWriterUtils.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/RowCSVWriterUtils.scala
@@ -36,8 +36,9 @@ object RowCSVWriterUtils {
       writer.writeStringField(json)
     }
   }
-  private def tss (l:Long, timeZone: ZoneId) ={
-    LocalDateTime.ofInstant(Instant.EPOCH.plus(l, ChronoUnit.MICROS), timeZone).toString
+
+  private def getLocalDateTimeFromTimestampWithZone (timestamp:Long, timeZone: ZoneId) ={
+    LocalDateTime.ofInstant(Instant.EPOCH.plus(timestamp, ChronoUnit.MICROS), timeZone)
   }
 
   // This method does not check for null at the current row idx and should be checked before !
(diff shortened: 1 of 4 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/datasink/RowCSVWriterUtils.scala
```diff
@@ -35,25 +36,30 @@ object RowCSVWriterUtils {
       writer.writeStringField(json)
     }
   }
+  private def tss (l:Long, timeZone: ZoneId) ={
+    LocalDateTime.ofInstant(Instant.EPOCH.plus(l, ChronoUnit.MICROS), timeZone).toString
```
yihezkel, line 40: I think this method would make more sense returning a LocalDateTime. The caller can call "toString" on it.
Plus, that would be consistent with the other "writer.writeStringField" calls below.

connector/src/test/scala/com/microsoft/kusto/spark/WriterTests.scala
```diff
@@ -142,8 +139,8 @@ class WriterTests extends FlatSpec with Matchers {
         "asd2" -> Row(Array("stringVal2\b\f")))
     )
 
-    // 2018-06-18 16:04:00.0 in GMT
-    val ts = new java.sql.Timestamp(1529337840000l)
+    val tz  = ZonedDateTime.parse("2018-06-18T15:00:00.123456789+04:00")
```
yihezkel, line 142: I could be missing it, but I don't see an assert that this or any non-UTC LocalDateTime went through the writer and came out the other end as expected.

connector/src/main/scala/com/microsoft/kusto/spark/datasink/RowCSVWriterUtils.scala
```diff
@@ -35,25 +36,30 @@ object RowCSVWriterUtils {
       writer.writeStringField(json)
     }
   }
+  private def tss (l:Long, timeZone: ZoneId) ={
```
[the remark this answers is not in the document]
ohadbitt: oo right - that was supposed to be temporary

connector/src/main/scala/com/microsoft/kusto/spark/datasink/RowCSVWriterUtils.scala
ohadbitt: thats right

connector/src/test/scala/com/microsoft/kusto/spark/WriterTests.scala
ohadbitt: line 282
it sohuld equal to 
sdf.format(ts.toInstant.atZone(tz.getZone))

yihezkel (approved):

PR #188, opened 2021-06-08
New option for the auto delete policy time - apart from the timeout w…

ohadbitt:
**Features:**
- New option for the auto delete policy time - apart from the timeout w…

**Fixes:**
- Users suddenly get their tables deleted in the middle of the flow

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoSinkOptions.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoSinkOptions.scala
@@ -33,6 +33,9 @@ object KustoSinkOptions extends KustoOptions{
   // partition. Kusto's ingestion also aggregates data, default suggested by Kusto is 1GB but here we suggest to cut
   // it at 100MB to adjust it to spark pulling of data.
   val KUSTO_CLIENT_BATCHING_LIMIT: String = newOption("clientBatchingLimit")
+
+  // Time after which staging resources are cleaned if they weren't cleaned at the end of the run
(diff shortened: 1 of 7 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoSinkOptions.scala
```diff
@@ -33,6 +33,9 @@ object KustoSinkOptions extends KustoOptions{
   // partition. Kusto's ingestion also aggregates data, default suggested by Kusto is 1GB but here we suggest to cut
   // it at 100MB to adjust it to spark pulling of data.
   val KUSTO_CLIENT_BATCHING_LIMIT: String = newOption("clientBatchingLimit")
+
+  // Time after which staging resources are cleaned if they weren't cleaned at the end of the run
```
yogilad, line 37: In seconds, minutes, is there a default?

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoSinkOptions.scala
```diff
@@ -33,6 +33,9 @@ object KustoSinkOptions extends KustoOptions{
...
+  val KUSTO_STAGING_RESOURCES_CLEANUP: String = newOption("stagingResourcesCleanup")
```
yogilad, line 38: The name does not imply the value should be a duration or that the option is a sort of timeout.
Maybe KUSTO_STAGING_RESOURCE_AUTO_CLEANUP_TIMEOUT?
(long but perhaps better)

yogilad (changes requested):

yogilad (approved):

PR #192, opened 2021-06-22
Extents move batches

ohadbitt:
Extents move by batches with retries


yihezkel (approved):
Comments as discussed on phone

PR #202, opened 2021-08-23
Add ClientRequstProperties to Write + transient errors check on extents move retry

ohadbitt:
ClientRequstProperties option can be used on write operations (used for schema check operation, table creation and move extents)
Transient errors check on extents move retry + specific handling of timeout until changed from SDK side

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
@@ -125,7 +125,7 @@ class KustoClient(val clusterAlias: String, val engineKcsb: ConnectionStringBuil
   }
 
   def findErrorInResult(res: KustoResultSetTable): (Boolean, Object) = {
-    var error:Object = null
+    var error: Object = null
     var failed = false
     if (KDSU.getLoggingLevel == Level.DEBUG) {
       var i = 0
@@ -154,8 +154,8 @@ class KustoClient(val clusterAlias: String, val engineKcsb: ConnectionStringBuil
     (failed, error)
   }
 
-  def moveExtentsWithRetries(batchSize: Int, totalAmount: Int, database: String, tmpTableName: String, targetTable: String, crp: ClientRequestProperties,
-                             writeOptions: WriteOptions): Unit = {
+  def moveExtentsWithRetries(batchSize: Int, totalAmount: Int, database: String, tmpTableName: String, targetTable: String,
+                             crp: ClientRequestProperties, writeOptions: WriteOptions): Unit = {
     var extentsProcessed = 0
     var retry = 0
     var curBatchSize = batchSize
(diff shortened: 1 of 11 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
```diff
@@ -118,67 +124,87 @@ class KustoClient(val clusterAlias: String, val engineKcsb: ConnectionStringBuil
     extentsLeftRes.getInt(0) != 0
   }
 
+  def findErrorInResult(res: KustoResultSetTable): (Boolean, Object) = {
+    var error:Object = null
+    var failed = false
+    if (KDSU.getLoggingLevel == Level.DEBUG) {
```
yogilad, line 130: Can you consolidate the loops so that debugging is an internal decision?

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
```diff
@@ -118,67 +124,87 @@ class KustoClient(val clusterAlias: String, val engineKcsb: ConnectionStringBuil
...
+      var i = 0
+      while (res.next() && !failed) {
+        val targetExtent = res.getString(1)
+        error = res.getObject(2)
+        if (targetExtent == "Failed" || StringUtils.isNotBlank(error.asInstanceOf[String])) {
+          failed = true
+          if (i > 0) {
+            KDSU.logFatal(myName, "Failed extent was not reported on all extents!." +
```
yogilad, line 138: This is enclosed in the Debug clause, yet seems like you expect customers to see and respond to this alert

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoDataSourceUtils.scala
```diff
@@ -248,35 +253,36 @@ object KustoDataSourceUtils {
...
-    if (crpOption.isDefined) {
-      val crp = ClientRequestProperties.fromString(crpOption.get)
-      crp
+    val crp = if (crpOption.isDefined) {
+      ClientRequestProperties.fromString(crpOption.get)
     } else {
       new ClientRequestProperties
     }
+
+    crp.setClientRequestId(requestId)
```
yogilad, line 284: Assuming the user has provider a requestId, do you allow it or want to override it?

yogilad (approved):

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
ohadbitt: My intention here was that as this was fast written and tested that if user complains he would turn the logging to debug and send us complete logs after that

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoDataSourceUtils.scala
ohadbitt: if user provided both requestId and ClientRequestProperties the requestId overrides the one in the crp

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
ohadbitt: what do you mean by that ?

yogilad: The _if_ and _else_ include share similar code apart for breaking on failures and reporting alerts.
Can you consolidate the code so If Debug is internalized?

You can send him a Jar with this, but it would be better for the official code to include this change

PR #203, opened 2021-09-03
Make it case sensitive when get partition columns

ivoson:
#### Pull Request Description
Make is case sensitive when get partition column, since kusto take schema as case sensitive if I understand this right.
If I specify a partition column while the schema itself has upper letters, the query will fail.

---

#### Future Release Comment
- None

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- None


ohadbitt:
We can do that - although partitioning is not supported - this was some triage and its on hold until kusto will be able to partition on export

PR #207, opened 2021-09-09
Async merge

ohadbitt:
Async merge so that it doesnt need to keep connection

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoReader.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoReader.scala
@@ -244,6 +244,7 @@ private[kusto] class KustoReader(client: Client, request: KustoReadRequest, stor
     val commandResult: KustoResultSetTable = client.execute(request.kustoCoordinates.database,
       exportCommand,
       request.clientRequestProperties.orNull).getPrimaryResults
-    KDSU.verifyAsyncCommandCompletion(client, request.kustoCoordinates.database, commandResult, directory,timeOut = request.timeout)
+    KDSU.verifyAsyncCommandCompletion(client, request.kustoCoordinates.database, commandResult, timeOut = request
+      .timeout, doingWhat = s"export data to  blob directory: ('$directory') preparing it for reading.")
   }
 }
--- a/connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
@@ -63,11 +63,23 @@ private[kusto] object CslCommandsGenerator {
        |  distinct ExtentId"""
   }
 
+  def generateTableMoveExtentsAsyncCommand(sourceTableName: String, destinationTableName: String, batchSize:Int): String
+  = {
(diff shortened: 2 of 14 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoReader.scala
```diff
@@ -244,6 +244,7 @@ private[kusto] class KustoReader(client: Client, request: KustoReadRequest, stor
     val commandResult: KustoResultSetTable = client.execute(request.kustoCoordinates.database,
       exportCommand,
       request.clientRequestProperties.orNull).getPrimaryResults
-    KDSU.verifyAsyncCommandCompletion(client, request.kustoCoordinates.database, commandResult, directory,timeOut = request.timeout)
+    KDSU.verifyAsyncCommandCompletion(client, request.kustoCoordinates.database, commandResult, timeOut = request
+      .timeout, doingWhat = s"export data to  blob directory: ('$directory') preparing it for reading.")
```
yogilad, line 248: nit: 
Extra space between "to" and "blob"
"in preparation for reading"?

connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
```diff
@@ -63,11 +63,23 @@ private[kusto] object CslCommandsGenerator {
        |  distinct ExtentId"""
   }
 
+  def generateTableMoveExtentsAsyncCommand(sourceTableName: String, destinationTableName: String, batchSize:Int): String
+  = {
+    s""".move async extents to table $destinationTableName <|
+       .show table $sourceTableName extents with(extentsShowFilteringRuntimePolicy='{"MaximumResultsCount":$batchSize}');
+        $$command_results
```
yogilad, line 70: What's command_results (not an argument like the rest)?

connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
```diff
@@ -63,11 +63,23 @@ private[kusto] object CslCommandsGenerator {
...
+       |  distinct ExtentId"""
```
yogilad, line 71: You'll get a list of only unique extent ids.
* Don't you need the operationIds?
* Why do you need to call distinct? where are the duplications?

connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
ohadbitt: command results is a KQL macro in command-then-query flow - to get the command results as a Table and operate on it

connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
ohadbitt: this is one operation 
i dont need distinct anymore - this command changed a while ago - will fix

yogilad (approved):
Check this well since I'm not the authority on the Java / Spark SDK.
You can also want to get another look from Asaf or Yihezkel

PR #216, opened 2022-01-10
Fix a few bugs that prevent compilation in Scala 2.11.11

yihezkel:
#### Pull Request Description

Fix a few bugs that prevent compilation in Scala 2.11.11

---

#### Future Release Comment
**Fixes:**
- Excluded many duplicate scala-library dependencies that may cause conflicts
- Added method to convert a scala lambda to a Java Consumer so we can both create a Consumer[DeviceCode] and also use closures to update the device code locally (not idea solution, but improving it is for another release).
- Convert Java list to a scala list so we can iterate over it

### DIFF
--- a/connector/pom.xml
+++ b/connector/pom.xml
@@ -213,6 +213,10 @@
                     <artifactId>jackson-module-scala_2.12</artifactId>
                     <groupId>com.fasterxml.jackson.module</groupId>
                 </exclusion>
+                <exclusion>
+                    <artifactId>scala-library</artifactId>
+                    <groupId>org.scala-lang</groupId>
+                </exclusion>
             </exclusions>
         </dependency>
         <dependency>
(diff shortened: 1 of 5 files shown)


ohadbitt:
I think this should be done to branch 2.4 and not master

yihezkel:
@ohadbitt 
Thanks, I didn't realize we managed our 2.4 code via that branch. Do you have any objection to me adding that to our README?

But also, I feel like if we have a choice between having a block of code that differs between 2.4 and 3.x, and changing that block in both so that the same code compiles in both, it'd be better to do the latter so that branches 2.4 and master diverge as little as possible.
WDYT?

ohadbitt:
> 

sounds good if you can do it

ohadbitt (approved):

PR #221, opened 2022-01-30
Add support for Materialized Views

yihezkel:
#### Pull Request Description

Add support for Materialized Views by using new SetNewIngestionTime feature in such cases
Breaking change: Renamed `KUSTO_MAX_RETRIES_ON_MOVR_EXTENTS` to `KUSTO_MAX_RETRIES_ON_MOVE_EXTENTS`

---

#### Future Release Comment
**Features:**
- If ingesting to a table which is the source for a Materialized View, append the newly added `SetNewIngestionTime=true` capability when moving the extents from the temp table.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,48 +8,48 @@ master: [![Build status](https://msazure.visualstudio.com/One/_apis/build/status
  
 This library contains the source code for Azure Data Explorer Data Source and Data Sink Connector for Apache Spark.
 
-Azure Data Explorer (A.K.A. [Kusto](https://azure.microsoft.com/en-us/services/data-explorer/)) is a lightning-fast indexing and querying service. 
+Azure Data Explorer (A.K.A. [Kusto](https://azure.microsoft.com/services/data-explorer/)) is a lightning-fast indexing and querying service. 
 
 [Spark](https://spark.apache.org/) is a unified analytics engine for large-scale data processing.
 
-Making Azure Data Explorer and Spark work together enables building fast and scalable applications, targeting a variety of Machine Learning, Extract-Transform-Load, Log Analytics and other data driven scenarios. 
+Making Azure Data Explorer and Spark work together enables building fast and scalable applications, targeting a variety of Machine Learning, Extract-Transform-Load, Log Analytics and other data-driven scenarios. 
 
 ## Changelog
 
-For main changes from previous releases please refer to [Releases](https://github.com/Azure/azure-kusto-spark/releases).
-For known or new issues please refer to the [issues](https://github.com/Azure/azure-kusto-spark/issues) section.
+For major changes from previous releases, please refer to [Releases](https://github.com/Azure/azure-kusto-spark/releases).
+For known or new issues, please refer to the [issues](https://github.com/Azure/azure-kusto-spark/issues) section.
 
 ## Usage
 
 ### Linking 
 
-Starting version 2.3.0 we introduce new artifact Ids: `kusto-spark_3.x_2.12` targeting Spark 3.x and Scala 2.12 
-and `kusto-spark_2.4_2.11` targeting Spark 2.4.x and scala 2.11.
+Starting version 2.3.0, we introduce new artifact Ids: `kusto-spark_3.x_2.12` targeting Spark 3.x and Scala 2.12 
+and `kusto-spark_2.4_2.11` targeting Spark 2.4.x and scala 2.11. The latter is built from branch "2.4".
 For Scala/Java applications using Maven project definitions, 
-link your application with the artifact below in order to use the Azure Data Explorer connector for Spark. 
-> Note: Versions prior to 2.5.1  do not work anymore for ingest to an existing table, please update to the latest.
+link your application with the artifact below to use the Azure Data Explorer Connector for Spark. 
+> Note: Versions prior to 2.5.1 can no longer ingest to an existing table.
 
 ```
 groupId = com.microsoft.azure.kusto
 artifactId = kusto-spark_3.0_2.12
-version = 2.9.1
+version = 2.9.3
 ```
 
 **In Maven**:
 
 Look for the following coordinates: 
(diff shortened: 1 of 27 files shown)


README.md
```diff
@@ -124,10 +124,10 @@ Usage examples can be found [here](samples/src/main/scala)
 
 # Available Azure Data Explorer client libraries:
 
-Here is a list of currently available client libraries for Azure Data Explorer:
+Currently available client libraries for Azure Data Explorer:
 - [Node](https://github.com/azure/azure-kusto-node)
 - [Python](https://github.com/azure/azure-kusto-python)
-- [.NET](https://docs.microsoft.com/en-us/azure/kusto/api/netfx/about-the-sdk)
+- [.NET](https://docs.microsoft.com/azure/kusto/api/netfx/about-the-sdk)
 - [Java](https://github.com/azure/azure-kusto-java)
```
AsafMah, line 131: Add go? https://github.com/azure/azure-kusto-go

connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
```diff
@@ -55,16 +56,22 @@ private[kusto] object CslCommandsGenerator {
     s".show export containers"
   }
 
-  def generateTableMoveExtentsCommand(sourceTableName: String, destinationTableName: String, batchSize:Int): String = {
-    s""".move extents to table $destinationTableName <|
+  def generateIsTableMaterializedViewSourceCommand(destinationTableName: String): String = {
+    s""".show materialized-views | where SourceTable == '$destinationTableName' | count"""
+  }
+
+  def generateTableMoveExtentsCommand(sourceTableName: String, destinationTableName: String, batchSize: Int, isDestinationTableMaterializedViewSource: Boolean = false): String = {
```
AsafMah, line 63: formatting suggestion:

```scala
  def generateTableMoveExtentsCommand(sourceTableName: String,
                                      destinationTableName: String,
                                      batchSize: Int,
                                      isDestinationTableMaterializedViewSource: Boolean = false): String = {
    val setNewIngestionTime: String = if (isDestinationTableMaterializedViewSource) "with(SetNewIngestionTime=true)" else ""
    
    s""".move extents to table $destinationTableName $setNewIngestionTime <|
       .show table $sourceTableName extents with(extentsShowFilteringRuntimePolicy='{"MaximumResultsCount":$batchSize}');
        $$command_results
       |  distinct ExtentId"""
    
  }

  def generateTableMoveExtentsAsyncCommand(sourceTableName: String,
                                           destinationTableName: String,
                                           batchSize: Int,
                                           isDestinationTableMaterializedViewSource: Boolean = false): String = {
    val setNewIngestionTime: String = if (isDestinationTableMaterializedViewSource) "with(SetNewIngestionTime=true)" else ""
    
    s""".move async extents to table $destinationTableName $setNewIngestionTime <|
       .show table $sourceTableName extents with(extentsShowFilteringRuntimePolicy='{"MaximumResultsCount":$batchSize}');
       """
  }
```

connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
```diff
@@ -55,16 +56,22 @@ private[kusto] object CslCommandsGenerator {
     s".show export containers"
   }
 
-  def generateTableMoveExtentsCommand(sourceTableName: String, destinationTableName: String, batchSize:Int): String = {
-    s""".move extents to table $destinationTableName <|
+  def generateIsTableMaterializedViewSourceCommand(destinationTableName: String): String = {
```
AsafMah, line 59: Do we have to do two roundtrips to kusto for this? Can't we put this result in a `let` binding, and check it within the command?

AsafMah (commented):
Looks nice, some comments though.

BTW while the formatting and typo fixes are very blessed, it was hard to find the "meat" of the code within it

connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
yihezkel: Not necessarily.

I'm very happy you reviewed this already, but the reason why I didn't yet request your review was because this is my first stab - I haven't even tested it! So this is the brute-force approach, and I intend to optimize once I test and debug.

Again - it's great to have your review even now before it's fully baked, but my intent was that this would be a "draft PR".

ohadbitt (approved):
Just make sure it works
See that you recreate the bug without and see it fixes

connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
yihezkel: Unfortunately, it doesn't seem admin commands against the engine can be part of `let` statements.

docs/Authentication.md
```diff
@@ -161,19 +161,19 @@ df.write
...
->**Note:** Device authentication is not recommended for production   
+>**Note:** Device authentication is not recommended for production
 
 ## Authentication via token provider callback
 
-User can provide a class to be used as a callback for providing an AAD token for authentication. 
+User can provide a class to be used as a callback for providing an AAD authentication token. 
 The class should have a constructor that accepts one argument of type CaseInsensitiveMap[String] (which will contain
-the options provided to the connector). The class should extend Callable<String> with Serializeable and is expected
+the options provided to the Connector). The class should extend Callable<String> with Serializeable and is expected
```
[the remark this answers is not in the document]
yihezkel: My thought process is based on the fact that "Connector" is short for "Azure Data Explorer Connector", which is a proper noun, and thus all parts are capitalized.

Good catch - it can go either way, though.

PR #232, opened 2022-03-23
Create temp table with same batching policy as destination table

yihezkel:
#### Pull Request Description

Create temp table with same batching policy as destination table. Call refresh cache DM command to refresh it.

---

#### Future Release Comment

**Features:**
- Support ingestion batching policy by creating temp table with same batching policy as destination table


yihezkel:
Sorry @ohadbitt about the confusion. I abandoned the previous PR because I somehow built that branch atop some 2.4 changes :-)

These are the changes I intended.

connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
```diff
@@ -27,6 +27,14 @@ private[kusto] object CslCommandsGenerator {
     s".drop tables ($tables) ifexists"
   }
 
+  def generateShowTableIngestionBatchingPolicyCommand(tableName: String): String = {
+    s".show table ${KustoQueryUtils.normalizeTableName(tableName)} policy ingestionbatching"
+  }
+
+  def generateAlterTableIngestionBatchingPolicyCommand(tableName: String, targetTableBatchingPolicy: String): String = {
+    s""".alter table ${KustoQueryUtils.normalizeTableName(tableName)} policy ingestionbatching @"$targetTableBatchingPolicy""""
```
yihezkel, line 35: Per discussion with @yogilad:
Nobody cares what the batching policy is of the temp table. Instead, we just want to make sure that the data gets to the destination table without an unexpected (e.g. 5m) delay. So we can set for the temp table a hardcoded very low batching policy (e.g. 5s).
Consider using the file size and number of file batching policies of the destination table (as I already do in this PR) so we don't create unoptimized extents, and only hardcode the time component.

The above assumes we only start this process after we reached the end of their data - confirm this.

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
```diff
@@ -82,6 +82,8 @@ class KustoClient(val clusterAlias: String, val engineKcsb: ConnectionStringBuil
     // Create a temporary table with the kusto or dataframe parsed schema with retention and delete set to after the
     // write operation times out. Engine recommended keeping the retention although we use auto delete.
     engineClient.execute(database, generateTempTableCreateCommand(tmpTableName, tmpTableSchema), crp)
+    val targetTableBatchingPolicy = "{'MaximumBatchingTimeSpan':'00:00:05', 'MaximumNumberOfItems': 100, 'MaximumRawDataSizeMB': 300}"
+    engineClient.execute(database, generateAlterTableIngestionBatchingPolicyCommand(tmpTableName, targetTableBatchingPolicy), crp)
```
yihezkel, line 86: `DmClusterOps` permission is required to execute `.refresh policy cache`

ohadbitt (approved):

PR #242, opened 2022-06-13
Fix bugs and upgrade to sdk v3.1.1

ohadbitt:
#### Future Release Comment

**Features:**
- Upgrade to kusto java sdk v3.1.1
- Fix device authentication
- Avoid deprecated mapping ingestion property
- Dont fail on alias extract
- Materialized view move extents flag should only be used on engine v3


connector/src/main/scala/com/microsoft/kusto/spark/authentication/DeviceAuthentication.scala
```diff
@@ -1,28 +1,25 @@
...
 import scala.concurrent.TimeoutException
 
 class DeviceAuthentication (val cluster: String, val authority:String) extends auth.DeviceAuthTokenProvider(cluster, authority) {
-  var deviceCode: Option[DeviceCode] = None
+  var currentDeviceCode: Option[DeviceCode] = None
   var expiresAt: Option[Long] = None
-  var awaitAuthentication: Option[CompletableFuture[IAuthenticationResult]] = None
-  val NewDeviceCodeFetchTimeout = 5000
-  val Interval = 500
+  val NewDeviceCodeFetchTimeout = 60L * 1000L
```
yihezkel, line 13: nit:lower case

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
```diff
@@ -164,25 +164,36 @@ class KustoClient(val clusterAlias: String, val engineKcsb: ConnectionStringBuil
...
+  }
+
   def moveExtentsWithRetries(batchSize: Int, totalAmount: Int, database: String, tmpTableName: String, targetTable: String,
                              crp: ClientRequestProperties, writeOptions: WriteOptions): Unit = {
     var extentsProcessed = 0
     var retry = 0
     var curBatchSize = batchSize
     var delayPeriodBetweenCalls = DelayPeriodBetweenCalls
     var consecutiveSuccesses = 0
+    val useMaterializedViewFlag = shouldUseMaterializedViewFlag(database, targetTable, crp)
```
yihezkel, line 188: Maybe you already did, but I think we should clarify with Yifat what to do if it is a MV, but we're on EngineV2. I would think we should then fail and tell the user it isn't supported.

yihezkel (approved):
Looks great! Just a few suggestions.

connector/src/main/scala/com/microsoft/kusto/spark/authentication/DeviceAuthentication.scala
ohadbitt: This is the scala convention for consts

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
ohadbitt: Talked to Yifat about it - and she suggested to let it be as it was

PR #243, opened 2022-06-21
Polling and Move extents as spark job

ohadbitt:
**Breaking Changes:**
- Currently users seem to think the kusto.write operation is hang after data processing reaches 100%, therefore we put both into new jobs that will be monitored by Spark and visualized to the user. This causes one executor thread of the worker nodes to become idle and therefore can process less jobs concurrently so for production purposes we suggest to turn this down using the new option.


connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoSinkOptions.scala
```diff
@@ -57,7 +61,8 @@ object SchemaAdjustmentMode extends Enumeration {
   val NoAdjustment, FailIfNotMatch, GenerateDynamicCsvMapping = Value
 }
 
-case class WriteOptions(tableCreateOptions: SinkTableCreationMode.SinkTableCreationMode = SinkTableCreationMode.FailIfNotExist,
+case class WriteOptions(pollingOnDriver:Boolean = true,
```
yihezkel, line 64: Shouldn't we tune the default for prod instead of dev?

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -93,9 +94,15 @@ object KustoWriter {
       KDSU.logInfo(myName, s"Successfully created temporary table $tmpTableName, will be deleted after completing the operation")
 
       kustoClient.setMappingOnStagingTableIfNeeded(stagingTableIngestionProperties, tableCoordinates.database, tmpTableName, table, crp)
-      if (stagingTableIngestionProperties.flushImmediately) {
-        KDSU.logWarn(myName, "It's not recommended to set flushImmediately to true")
+//    TODO remove until batching policy problem is good
+
```
yihezkel, line 98: nit: Move space to before the TODO

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -93,9 +94,15 @@ object KustoWriter {
       KDSU.logInfo(myName, s"Successfully created temporary table $tmpTableName, will be deleted after completing the operation")
 
       kustoClient.setMappingOnStagingTableIfNeeded(stagingTableIngestionProperties, tableCoordinates.database, tmpTableName, table, crp)
-      if (stagingTableIngestionProperties.flushImmediately) {
-        KDSU.logWarn(myName, "It's not recommended to set flushImmediately to true")
+//    TODO remove until batching policy problem is good
```
yihezkel, line 97: Why? Isn't `flushImmediately` still useful?

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
```diff
@@ -289,74 +291,99 @@ class KustoClient(val clusterAlias: String, val engineKcsb: ConnectionStringBuil
...
+                KDSU.logInfo(loggerName, s"Ingestion to Kusto succeeded. " +
+                  s"Cluster: '${coordinates.clusterAlias}', " +
+                  s"database: '${coordinates.database}', " +
+                  s"table: '$tmpTableName'$batchIdIfExists, partition: '${partitionResult.partitionId}'', from: '${finalRes.get.ingestionSourcePath}', Operation ${finalRes.get.operationId}")
+              case otherStatus =>
+
+                throw new RuntimeException(s"Ingestion to Kusto failed with status '$otherStatus'." +
+                  s" Cluster: '${coordinates.clusterAlias}', database: '${coordinates.database}', " +
+                  s"table: '$tmpTableName'$batchIdIfExists, partition: '${partitionResult.partitionId}'." +
+                  "")//s" Ingestion info: '${readIngestionResult(finalRes.get)}'")
```
yihezkel, line 342: ?

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
ohadbitt: It is 
I removed the warning as it seems that as long as we don't have   a solution with batching policy it is a good practice to use it

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoSinkOptions.scala
ohadbitt: We want first time users to understand what is happening and have this experience

yihezkel: I personally think the defaults should be for prod, and the README instructs the need to temporarily adjust for dev.

yihezkel (approved):

PR #244, opened 2022-07-05
Bug fix : Dont swallow timeout exception

ohadbitt:
Also increased timeout to 30 seconds

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
@@ -252,7 +252,6 @@ object KustoWriter {
         val retry = Retry.of("Ingest to Kusto", this.retryConfig)
         val retryExecute: CheckedFunction0[IngestionResult] = Retry.decorateCheckedSupplier(retry, () => {
           KDSU.logInfo(myName, s"Queued blob for ingestion in partition $partitionId for requestId: '$requestId}")
-          if(partitionId==1) throw new IngestionServiceException("po")
           ingestClient.ingestFromBlob(blobSourceInfo, props)
         })
         partitionsResults.add(PartitionResult(retryExecute.apply,partitionId))
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
@@ -1,15 +1,8 @@
 package com.microsoft.kusto.spark.datasink
 
-import java.io._
-import java.net.URI
-import java.nio.charset.StandardCharsets
-import java.security.InvalidParameterException
-import java.util
-import java.util.zip.GZIPOutputStream
-import java.util.{TimeZone, UUID}
 import com.microsoft.azure.kusto.data.ClientRequestProperties
 import com.microsoft.azure.kusto.ingest.IngestionProperties.DataFormat
-import com.microsoft.azure.kusto.ingest.exceptions.IngestionClientException
+import com.microsoft.azure.kusto.ingest.exceptions.IngestionServiceException
 import com.microsoft.azure.kusto.ingest.result.IngestionResult
 import com.microsoft.azure.kusto.ingest.source.BlobSourceInfo
 import com.microsoft.azure.kusto.ingest.{IngestClient, IngestionProperties}
@@ -29,7 +22,13 @@ import org.apache.spark.sql.types.StructType
 import org.apache.spark.util.CollectionAccumulator
 import org.json.JSONObject
 
-import java.util.concurrent.TimeUnit
+import java.io._
+import java.net.URI
+import java.nio.charset.StandardCharsets
+import java.security.InvalidParameterException
+import java.util
+import java.util.zip.GZIPOutputStream
+import java.util.{TimeZone, UUID}
 import scala.collection.Iterator
 import scala.collection.JavaConverters._
 import scala.concurrent.ExecutionContext.Implicits.global
@@ -251,7 +250,8 @@ object KustoWriter {
           KDSU.logInfo(myName, s"Queued blob for ingestion in partition ${TaskContext.getPartitionId} for requestId: '$requestId}")
           ingestClient.ingestFromBlob(blobSourceInfo, props)
         } catch {
-          case _: IngestionClientException =>
+          // Catch only IngestionServiceException cause it can be due to transient storage issue
+          case _: IngestionServiceException =>
             ingestClient.ingestFromBlob(blobSourceInfo, props)
         }
         partitionsResults.add(PartitionResult(res,partitionId))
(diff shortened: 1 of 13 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -192,7 +198,11 @@ object KustoWriter {
     val partitionId = TaskContext.getPartitionId
     KDSU.logInfo(myName, s"Ingesting partition: '$partitionId' in requestId: '${writeOptions.requestId}'$batchIdForTracing")
     val ingestClient = KustoClientCache.getClient(coordinates.clusterAlias, coordinates.clusterUrl, authentication).ingestClient
```
yogilad, line 200: Feels like this sets the options multiple times on the same client.
Should it not be done once when the client is added to the cache?

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -237,7 +247,14 @@ object KustoWriter {
         val blobPath = blobUri + sas
         val blobSourceInfo = new BlobSourceInfo(blobPath, size)
 
-        partitionsResults.add(PartitionResult(ingestClient.ingestFromBlob(blobSourceInfo, props), partitionId))
+        val res: IngestionResult = try {
+          KDSU.logInfo(myName, s"Queued blob for ingestion in partition ${TaskContext.getPartitionId} for requestId: '$requestId}")
+          ingestClient.ingestFromBlob(blobSourceInfo, props)
+        } catch {
+          case _: IngestionClientException =>
+            ingestClient.ingestFromBlob(blobSourceInfo, props)
```
yogilad, line 255: Is this a retry?
Why this particular set exceptions?

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoConstants.scala
```diff
@@ -25,4 +25,8 @@ object KustoConstants {
   val DefaultBatchingLimit: Int = 100
   val DefaultExtentsCountForSplitMergePerNode: Int = 400
   val DefaultMaxRetriesOnMoveExtents: Int = 10
+  val DefaultExecutionQueueing: Int = TimeUnit.SECONDS.toMillis(15).toInt
```
yogilad, line 28: Is this 15msec or 15000msec?
(Sorry, not my native language)

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -237,7 +247,14 @@ object KustoWriter {
         val blobPath = blobUri + sas
         val blobSourceInfo = new BlobSourceInfo(blobPath, size)
 
-        partitionsResults.add(PartitionResult(ingestClient.ingestFromBlob(blobSourceInfo, props), partitionId))
+        val res: IngestionResult = try {
```
yogilad, line 250: Mind walking me through this section

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
ohadbitt: This will require some change but basically - getclient also add to cache if missing

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
ohadbitt: single retry

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
ohadbitt: will add a comment

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoConstants.scala
ohadbitt: 15000msec

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -155,9 +159,10 @@ object KustoWriter {
     KDSU.logInfo(myName, s"Ingesting from blob - partition: ${TaskContext.getPartitionId()} requestId: '${writeOptions.requestId}' $batchIdForTracing")
 
     tasks.asScala.foreach(t => try {
-      Await.result(t, KCONST.DefaultIngestionTaskTime)
+      Await.result(t, KCONST.DefaultMaximumIngestionTime)
+      // ALTHOUGH WE SAW THAT THE DM DIDNT EVEN GET THE MESSAGE EVENTUALLY
```
yihezkel, line 163: This looks like a note-to-self more than a comment intended for the public to see.

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -237,7 +246,15 @@ object KustoWriter {
         val blobPath = blobUri + sas
         val blobSourceInfo = new BlobSourceInfo(blobPath, size)
 
-        partitionsResults.add(PartitionResult(ingestClient.ingestFromBlob(blobSourceInfo, props), partitionId))
+        val res: IngestionResult = try {
+          KDSU.logInfo(myName, s"Queued blob for ingestion in partition ${TaskContext.getPartitionId} for requestId: '$requestId}")
+          ingestClient.ingestFromBlob(blobSourceInfo, props)
+        } catch {
```
yihezkel, line 252: Just thinking aloud:
This approach seems best if we will only ever want 1 retry and only if it's a `IngestionServiceException`. But I'm thinking we may want to add more in the future, and I can imagine wanting to retry more than once if for certain types of transient errors (like throttling). OTOH, we don't want to overengineer for a case that isn't relevant.

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -237,7 +246,15 @@ object KustoWriter {
...
+          // Catch only IngestionServiceException cause it can be due to transient storage issue
+          case _: IngestionServiceException =>
+            ingestClient.ingestFromBlob(blobSourceInfo, props)
```
yihezkel, line 255: Just making sure this was intentional - this is outside the try, so if it throws, it won't be added to `partitionsResults`.

yihezkel (approved):
Approved with suggestions

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
ohadbitt: It shouldnt - if we fail here - we fail everrything

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
ohadbitt: IngestionServiceException are all the transient errors from the service - we can rely on this conception it is - we might want to add this to documentation and code description but this is true

PR #247, opened 2022-07-20
Ingestion URI new option

ohadbitt:
New option to set ingestion uri - in case it cannot be deduced from engine url

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
@@ -4,18 +4,23 @@ import java.util.Locale
 
 trait KustoOptions {
   private val kustoOptionNames = collection.mutable.Set[String]()
-
   protected def newOption(name: String): String = {
     kustoOptionNames += name.toLowerCase(Locale.ROOT)
     name
   }
 
-  // KeyVault options. Relevant only if credentials need to be retrieved from Key Vault
-  val KEY_VAULT_URI = "keyVaultUri"
-  val KEY_VAULT_APP_ID = "keyVaultAppId"
-  val KEY_VAULT_APP_KEY = "keyVaultAppKey"
+  /** Required options */
+  // Target/source Kusto cluster for writing/reading the data.
+  val KUSTO_CLUSTER: String = newOption("kustoCluster")
 
+  // Target/source Kusto database for writing/reading the data. See KustoSink.md/KustoSource.md for
+  // required permissions
+  val KUSTO_DATABASE: String = newOption("kustoDatabase")
 
+  /**
+   * Authentication parameters - If non provided - device code would show on console. If console is not available
+   * use DeviceAuthentication class-https://github.com/Azure/azure-kusto-spark/blob/56667c54dfd43455113a9b48725b236a8d92ccd4/samples/src/main/python/pyKusto.py#L147
+   */
   // AAD application identifier of the client
   val KUSTO_AAD_APP_ID: String = newOption("kustoAadAppId")
   // AAD authentication authority
@@ -32,23 +37,26 @@ trait KustoOptions {
   // AAD application certificate password
   val KUSTO_AAD_APP_CERTIFICATE_PASSWORD: String = newOption("kutoAadAppCertPassword")
 
+  // KeyVault options. Relevant only if credentials need to be retrieved from Key Vault
+  val KEY_VAULT_URI = "keyVaultUri"
+  val KEY_VAULT_APP_ID = "keyVaultAppId"
+  val KEY_VAULT_APP_KEY = "keyVaultAppKey"
+
   // Classpath to a class that its constructor accepts one argument of type CaseInsensitiveMap[String] which will contain
   // the options provided to the connector. The class should extend Callable[String] with Serializeable and is expected
   // to return an AAD token upon invoking the call method.
   // The provider will be called for every request to the kusto service
   val KUSTO_TOKEN_PROVIDER_CALLBACK_CLASSPATH: String = newOption("tokenProviderCallbackClasspath")
 
-  // Target/source Kusto cluster for writing/reading the data.
-  val KUSTO_CLUSTER: String = newOption("kustoCluster")
-  // Target/source Kusto database for writing/reading the data. See KustoSink.md/KustoSource.md for
-  // required permissions
-  val KUSTO_DATABASE: String = newOption("kustoDatabase")
+  /** Optional parameters */
+  // Kusto ingestion uri cluster for reading data - provide this if ingestion uri cannot be deducted from adding
+  // "ingest-" prefix to the KUSTO_CLUSTER provided. i.e when using proxies.
+  val KUSTO_INGESTION_URI: String = newOption("kustoIngestionUri")
 
   // An integer number corresponding to the period in seconds after which the operation will timeout.
   // For write operations this limit starts ticking only after the data was processed by the connector and it starts
   // polling on the ingestion results. Default 2 days
   val KUSTO_TIMEOUT_LIMIT: String = newOption("timeoutLimit")
-
 
   // A json representation of the ClientRequestProperties object used for reading from Kusto
   var KUSTO_CLIENT_REQUEST_PROPERTIES_JSON: String = newOption("clientRequestPropertiesJson")
@@ -61,7 +69,8 @@ trait KustoOptions {
 case class KustoCoordinates(clusterUrl: String,
                             clusterAlias: String,
                             database: String,
-                            table: Option[String] = None)
+                            table: Option[String] = None,
+                            ingestionUrl: Option[String] = None)
 
 /** ******************************************************************************* */
 /*                                    NOTE!!!                                       */
--- a/connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
@@ -30,7 +30,7 @@ trait KustoOptions {
 
   val KUSTO_ACCESS_TOKEN: String = newOption("accessToken")
 
-  // Use only for local runs where one can open the browser
+  // Use only for local runs where one can open the browser - set to true to enable
   val KUSTO_USER_PROMPT: String = newOption("userPrompt")
   // AAD application pfx certificate path
   val KUSTO_AAD_APP_CERTIFICATE_PATH: String = newOption("kutoAadAppCertPath")
--- a/connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
@@ -18,7 +18,7 @@ trait KustoOptions {
   val KUSTO_DATABASE: String = newOption("kustoDatabase")
 
   /**
-   * Authentication parameters - If non provided - device code would show on console. If console is not available
+   * Authentication parameters - If not provided - device code would show on console. If console is not available
    * use DeviceAuthentication class-https://github.com/Azure/azure-kusto-spark/blob/56667c54dfd43455113a9b48725b236a8d92ccd4/samples/src/main/python/pyKusto.py#L147
    */
   // AAD application identifier of the client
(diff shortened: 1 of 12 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
```diff
@@ -4,18 +4,23 @@ import java.util.Locale
...
+  // Target/source Kusto cluster for writing/reading the data.
+  val KUSTO_CLUSTER: String = newOption("kustoCluster")
 
+  // Target/source Kusto database for writing/reading the data. See KustoSink.md/KustoSource.md for
+  // required permissions
+  val KUSTO_DATABASE: String = newOption("kustoDatabase")
 
+  /**
+   * Authentication parameters - If non provided - device code would show on console. If console is not available
+   * use DeviceAuthentication class-https://github.com/Azure/azure-kusto-spark/blob/56667c54dfd43455113a9b48725b236a8d92ccd4/samples/src/main/python/pyKusto.py#L147
```
yihezkel, line 22: Better to provide link to master branch (https://github.com/Azure/azure-kusto-spark/blob/master/samples/src/main/python/pyKusto.py)

connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
```diff
@@ -4,18 +4,23 @@ import java.util.Locale
...
+  /** Required options */
+  // Target/source Kusto cluster for writing/reading the data.
+  val KUSTO_CLUSTER: String = newOption("kustoCluster")
 
+  // Target/source Kusto database for writing/reading the data. See KustoSink.md/KustoSource.md for
+  // required permissions
+  val KUSTO_DATABASE: String = newOption("kustoDatabase")
 
+  /**
+   * Authentication parameters - If non provided - device code would show on console. If console is not available
```
yihezkel, line 21: non -> not

connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
```diff
@@ -25,31 +30,34 @@ trait KustoOptions {
...
-  // Use only for local runs where one can open the browser
+  // Use only for local runs where one can open the browser - set to true to enable
   val KUSTO_USER_PROMPT: String = newOption("userPrompt")
   // AAD application pfx certificate path
   val KUSTO_AAD_APP_CERTIFICATE_PATH: String = newOption("kutoAadAppCertPath")
   // AAD application certificate password
   val KUSTO_AAD_APP_CERTIFICATE_PASSWORD: String = newOption("kutoAadAppCertPassword")
 
+  // KeyVault options. Relevant only if credentials need to be retrieved from Key Vault
+  val KEY_VAULT_URI = "keyVaultUri"
```
yihezkel, line 41: Too late to change this now, but for the record this is a URL. URI is overly broad.

connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
```diff
@@ -25,31 +30,34 @@ trait KustoOptions {
...
 
-  // Target/source Kusto cluster for writing/reading the data.
-  val KUSTO_CLUSTER: String = newOption("kustoCluster")
-  // Target/source Kusto database for writing/reading the data. See KustoSink.md/KustoSource.md for
-  // required permissions
-  val KUSTO_DATABASE: String = newOption("kustoDatabase")
+  /** Optional parameters */
+  // Kusto ingestion uri cluster for reading data - provide this if ingestion uri cannot be deducted from adding
+  // "ingest-" prefix to the KUSTO_CLUSTER provided. i.e when using proxies.
+  val KUSTO_INGESTION_URI: String = newOption("kustoIngestionUri")
```
yihezkel, line 54: This is a URL

connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
```diff
@@ -25,31 +30,34 @@ trait KustoOptions {
...
   // The provider will be called for every request to the kusto service
   val KUSTO_TOKEN_PROVIDER_CALLBACK_CLASSPATH: String = newOption("tokenProviderCallbackClasspath")
 
-  // Target/source Kusto cluster for writing/reading the data.
-  val KUSTO_CLUSTER: String = newOption("kustoCluster")
-  // Target/source Kusto database for writing/reading the data. See KustoSink.md/KustoSource.md for
-  // required permissions
-  val KUSTO_DATABASE: String = newOption("kustoDatabase")
+  /** Optional parameters */
+  // Kusto ingestion uri cluster for reading data - provide this if ingestion uri cannot be deducted from adding
```
yihezkel, line 52: "Kusto ingestion cluster URL for reading data - provide this if ingestion URL cannot be deduced from adding"

connector/src/test/scala/com/microsoft/kusto/spark/KustoClientTests.scala
```diff
@@ -15,7 +15,8 @@ import scala.collection.JavaConverters._
 
 @RunWith(classOf[JUnitRunner])
 class KustoClientTests extends FlatSpec  with Matchers{
-  class KustoClientStub (override  val clusterAlias: String, override  val engineKcsb: ConnectionStringBuilder, override val ingestKcsb: ConnectionStringBuilder, var tagsToReturn: util.ArrayList[String]) extends KustoClient(clusterAlias, engineKcsb, ingestKcsb){
+  class KustoClientStub (override  val clusterAlias: String, override  val engineKcsb: ConnectionStringBuilder, override val ingestKcsb: ConnectionStringBuilder, var tagsToReturn: util.ArrayList[String])
+    extends KustoClient(engineKcsb, ingestKcsb, clusterAlias){
```
yihezkel, line 19: Is this standard to put the `extends` on a new line? It might be NBD, but I'm concerned it might make it harder to search the codebase.

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClientCache.scala
```diff
@@ -10,70 +10,67 @@ import com.microsoft.kusto.spark.utils.{KustoConstants => KCONST}
...
 
     engineKcsb.setClientVersionForTracing(KCONST.ClientName)
     ingestKcsb.setClientVersionForTracing(KCONST.ClientName)
 
-    new KustoClient(aliasAndAuth.clusterAlias, engineKcsb, ingestKcsb)
+    new KustoClient(engineKcsb, ingestKcsb, clusterAndAuth.clusterAlias)
   }
 
-  private[kusto] case class AliasAndAuth(clusterAlias: String, engineUrl: String, authentication: KustoAuthentication) {
+  private[kusto] case class ClusterAndAuth(engineUrl: String, authentication: KustoAuthentication, ingestionUri: Option[String], clusterAlias:String) {
```
yihezkel, line 63: I see you changed the order of parameters for some methods in this PR. Why not here as well?

I'd think it should be something like:
clusterAlias:String, authentication: KustoAuthentication, engineUrl: String, ingestionUri: Option[String]

connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
ohadbitt: none?

connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
ohadbitt: I really prefer not to - this way i can provide link to specific line

connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
ohadbitt: we call it URI everywhere

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClientCache.scala
ohadbitt: This is the main class that caused this change - i wanted alias to be last as it is less important

connector/src/test/scala/com/microsoft/kusto/spark/KustoClientTests.scala
ohadbitt: not really its actually used in new line only if its after last input variable

connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
ohadbitt: again - true - but this is not how it is called everywhere ..

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClientCache.scala
ohadbitt: so its actually did change accordingly

connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
yihezkel: You can do the same thing with the master branch:
https://github.com/Azure/azure-kusto-spark/blob/master/samples/src/main/python/pyKusto.py#L147

connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
yihezkel: I'd argue that we inconsistently vacillate between URL and URI:
image

In which case when creating a new variable, might as well do it right IMHO.

connector/src/test/scala/com/microsoft/kusto/spark/KustoClientTests.scala
yihezkel: I think it's better to keep it on the same line to make it easier to find, but your call.

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClientCache.scala
yihezkel: `alias` can be last, but I think it'd be best to improve the order. Specifically, `engineUrl: String` should be adjacent to `ingestionUri: Option[String]`

connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
ohadbitt: the difference is - this doesnt gurantee the line will stay on the same line

connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
ohadbitt: Maybe its a PM decision
I prefer to be concise with docs and other noteable instances i looked

connector/src/test/scala/com/microsoft/kusto/spark/KustoClientTests.scala
ohadbitt: yea i fixed

connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
yihezkel: Good point!

connector/src/main/scala/com/microsoft/kusto/spark/common/KustoOptions.scala
yihezkel: Should I ask the PM, or will you?

If you do it, I suggest including the difference between the 2 in your question. For example:
From https://www.geeksforgeeks.org/difference-between-url-and-uri, "URL is used to describe the identity of an item. URI provides a technique for defining the identity of an item. URL links a web page, a component of a web page or a program on a web page with the help of accessing methods like protocols. URI is used to distinguish one resource from other regardless of the method used. URL is a type of URI. URI is the superset of URL. URL is comprised of protocol, domain, path, hash, and so on. URI is comprised of scheme, authority, path, query and many more. Example URL: https://www.geeksforgeeks.org/ Example URI: isbn:0-294-56559-3"

I think that's important to be consistent, but I think it's more important to be accurate. Here, "URL" is more accurate (and consistent as well because we use both "URL" and "URI" terminology).

yihezkel (approved):

PR #256, opened 2022-08-22
Retry all the commands against the service

ohadbitt:


connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -287,16 +288,12 @@ object KustoWriter {
...
-        val f:CheckedFunction0[IngestionResult] = new CheckedFunction0[IngestionResult]() {
-          override def apply(): IngestionResult= {
-            KDSU.logInfo(myName, s"Queued blob for ingestion in partition $partitionId for requestId: '$requestId}")
-            ingestClient.ingestFromBlob(blobSourceInfo, props)
-          }
-        }
-
-        val retryExecute: CheckedFunction0[IngestionResult] = Retry.decorateCheckedSupplier(retry, f)
-        if (transactional) partitionsResults.add(PartitionResult(retryExecute.apply,partitionId))
+        if (transactional) partitionsResults.add(
```
yihezkel, line 291: I think this would be easier to read formatted as:
if (transactional) { // Even though body is technically 1 line
   // Remainder
}

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
```diff
@@ -48,6 +40,21 @@ class KustoClient(val engineKcsb: ConnectionStringBuilder, val ingestKcsb: Conne
     generateCreateTmpStorageCommand(), ingestProviderEntryCreator)
   private lazy val exportContainersContainerProvider = new ContainerProvider(dmClient, clusterAlias,
     generateGetExportContainersCommand(), exportProviderEntryCreator)
+  RetryConfig.ofDefaults()
+  private val BaseInterval = 1000L
```
yihezkel, line 44: BaseIntervalMs
MaxRetryIntervalMs

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
```diff
@@ -365,6 +372,16 @@ class KustoClient(val engineKcsb: ConnectionStringBuilder, val ingestKcsb: Conne
 
     shouldIngest
   }
+
+  def executeEngine(database: String, command: String, crp: ClientRequestProperties, retryConfig: Option[RetryConfig] = None): KustoOperationResult = {
+    KDSU.retryFunction(() => engineClient.execute(database, command, crp), retryConfig.getOrElse(this.retryConfig),
```
yihezkel, line 377: I find this (existing) design a bit strange. `engineClient`, `dmClient` and `ingestClient` are all basically singletons per KCSB that live as members within `KustoClient` - a class that I'd expect those 3 to be an **instance** of.

I'm sure there was wisdom to it beyond what I see in snippets here, and even if not, there isn't really anything to change about it here. So just calling it out, and asking if you know of wisdom behind it.

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
```diff
@@ -365,6 +372,16 @@ class KustoClient(val engineKcsb: ConnectionStringBuilder, val ingestKcsb: Conne
...
+      "Execute engine command with retries")
+  }
+
+  def executeDM(command: String, crp: ClientRequestProperties, retryConfig: Option[RetryConfig] = None): KustoOperationResult = {
+    KDSU.retryFunction(() => dmClient.execute("NetDefaultDB", command, crp), retryConfig.getOrElse(this.retryConfig),
```
yihezkel, line 382: "NetDefaultDB" should be a constant

yihezkel (approved):
Approved, with a few small comments

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClient.scala
ohadbitt: Decided on renaming 
With a suggestion for the future to split into three wrapper classes and a holder class

Issue #261, opened 2022-09-23
Fallback to export mode only if count query fails


ag-ramachandran:
The following is the behavior in case readMode is specified

a) If the user does not specify readMode , a count estimate is done and based on the ADX query limits there is switch to ForceDistributedMode (which uses export) or ForceSingleMode (which uses query directly). In the case that this count estimate query fails (query : Table | count), the fallback is to use **ForceDistributedMode**

b) If the user specifies the readMode as **ForceSingleMode** the estimate of the query limits is not performed. If the query fails in single mode , it will still fallback to export mode (ForceDistributedMode)

**Expected behavior**

The fallback to export mode **should not happen** in case **ForceSingleMode** is specified and the query fails.

PR #263, opened 2022-09-28
Fix Issue #261

ag-ramachandran:
Changes:

When creating buildScan operation, if the user explicitly uses the ForceSingleMode operation , always honor the single mode and do not fallback to distributed mode. This means that if ForceSingleMode is used and the query fails or times out the source operation will fail without fallback

In the earlier revision, if the ForcesingleMode is specified and the query fails, it will fallback to the ForceDistributedMode

Fixes issue #261

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
@@ -104,8 +104,8 @@ private[kusto] case class KustoRelation(kustoCoordinates: KustoCoordinates,
             } else {
               // The case where used did not provide an option and we estimated to be a single scan.
               // Our approximate estimate failed here , fallback to distributed
-              KDSU.logError("KustoRelation", "Failed with Single mode, falling back to Distributed mode," +
-                s"requestId: $requestId. Exception : ${exception.getMessage}")
+              KDSU.reportExceptionAndThrow("KustoRelation", exception,"Failed with Single mode, falling back to Distributed mode,",
+                kustoCoordinates.clusterAlias, kustoCoordinates.database, requestId = requestId, shouldNotThrow = true)
               readOptions.partitionOptions.column = Some(getPartitioningColumn)
               readOptions.partitionOptions.mode = Some(getPartitioningMode)
               KustoReader.distributedBuildScan(
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
@@ -4,7 +4,8 @@ import com.microsoft.azure.kusto.data.ClientRequestProperties
 import com.microsoft.kusto.spark.authentication.KustoAuthentication
 import com.microsoft.kusto.spark.common.{KustoCoordinates, KustoDebugOptions}
 import com.microsoft.kusto.spark.datasink.{KustoWriter, WriteOptions}
-import com.microsoft.kusto.spark.utils.{KustoClientCache, KustoConstants, KustoQueryUtils, KustoDataSourceUtils => KDSU}
+import com.microsoft.kusto.spark.utils.{ExtendedKustoClient, KustoClientCache, KustoConstants, KustoQueryUtils, KustoDataSourceUtils => KDSU}
+import org.apache.commons.lang3.exception.ExceptionUtils
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.sources._
 import org.apache.spark.sql.types.StructType
@@ -49,14 +50,20 @@ private[kusto] case class KustoRelation(kustoCoordinates: KustoCoordinates,
     val kustoClient = KustoClientCache.getClient(kustoCoordinates.clusterUrl, authentication,
       kustoCoordinates.ingestionUrl, kustoCoordinates.clusterAlias)
     val isUserOptionForceSingleMode = readOptions.readMode.contains(ReadMode.ForceSingleMode)
-    val storageParameters = maybeStorageParameters.getOrElse(kustoClient.getTempBlobsForExport)
     val (useSingleMode, estimatedRecordCount) = readOptions.readMode match {
       // if the user provides a specific option , this is to be used no matter what
       case Some(_) => (isUserOptionForceSingleMode, -1)
       // If there is no option mentioned , then estimate which option to use
       // Count records and see if we wat a distributed or single mode
-      case None => val estimatedRecordCountResult =  Try(KDSU.
-        estimateRowsCount(kustoClient.engineClient, query, kustoCoordinates.database, clientRequestProperties.orNull))
+      case None =>
+        val estimatedRecordCountResult = Try(KDSU.
+          estimateRowsCount(kustoClient.engineClient, query, kustoCoordinates.database, clientRequestProperties.orNull))
+        /*
+        Return values of estimate method
+        - Non zero integer : In case estimate method returns a value , estimated result
+        - Zero : Estimate fails and falls back to count , this gives a 0 result
+        - An exception : Happens when estimate is a null/empty and we fallback to count and count fails as well (timeout)
+        */
         estimatedRecordCountResult match {
           // if the count is lss than the 5k threshold,use single mode.
           case Success(recordCount) => (recordCount <= KustoConstants.DirectQueryUpperBoundRows, recordCount)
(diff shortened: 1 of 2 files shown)


ag-ramachandran:
Apart from the core method that needed change (buildscan) , the 2 accessor methods have been refactored to avoid lint errors and make them consistent with scala

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
```diff
@@ -38,79 +38,77 @@ private[kusto] case class KustoRelation(kustoCoordinates: KustoCoordinates,
...
-      if (count == 0 && !timedOutCounting) {
-        res = Some(sparkSession.emptyDataFrame.rdd)
-      } else {
-        // Use distributed mode if count is high or in case of a time out
-        useSingleMode =  !(timedOutCounting || count > KustoConstants.DirectQueryUpperBoundRows)
-      }
+    val kustoClient = KustoClientCache.getClient(kustoCoordinates.clusterUrl, authentication,
+      kustoCoordinates.ingestionUrl, kustoCoordinates.clusterAlias)
+    val isUserOptionForceSingleMode = readOptions.readMode.contains(ReadMode.ForceSingleMode)
+    val storageParameters = maybeStorageParameters.getOrElse(kustoClient.getTempBlobsForExport)
```
ohadbitt, line 52: please dont call kustoClient.getTempBlobsForExport if SingleMode is chosen (by user or us)

ohadbitt:
Please call these by the real change and add a description of what the change is and only add that it solves the issue at last line of the description. Also if you use '#' before the number it will link to the issue itself

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
```diff
@@ -38,79 +38,77 @@ private[kusto] case class KustoRelation(kustoCoordinates: KustoCoordinates,
...
-      if (useSingleMode) {
-        try {
-          res = Some(KustoReader.singleBuildScan(
-            kustoClient,
-            KustoReadRequest(sparkSession, cachedSchema, kustoCoordinates, query, authentication, timeout, clientRequestProperties, requestId),
-            KustoFiltering(requiredColumns, filters)))
-        } catch {
-          case ex: Exception => exception = Some(ex)
+    // if the selection yields 0 rows there are no records to scan and process
+    if(estimatedRecordCount == 0){
```
ohadbitt, line 68: This alarmed me for a second as i forgot my logic myself ...

Please comment it that this is zero only if a real count() kql was issue, if estimation returns zero or null estimatedRecordCount will equal 1

ag-ramachandran:
> Please call these by the real change and add a description of what the change is and only add that it solves the issue at last line of the description. Also if you use '#' before the number it will link to the issue itself

My bad, added a description of the behavior and linked the issue.

ohadbitt:
lets add to the documentation that we suggest to provide a value here as estimation can take up to 6 seconds

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
```diff
@@ -38,79 +38,77 @@ private[kusto] case class KustoRelation(kustoCoordinates: KustoCoordinates,
...
+      sparkSession.emptyDataFrame.rdd
+    } else {
+      // either a case of non-zero records or a case of timed-out.`
```
ohadbitt, line 71: Capital letter first comment word

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
```diff
@@ -38,79 +38,77 @@ private[kusto] case class KustoRelation(kustoCoordinates: KustoCoordinates,
...
+      if(useSingleMode){
+        // there are less than 5000 (KustoConstants.DirectQueryUpperBoundRows) records, perform a single scan
+        Try(KustoReader.singleBuildScan(
+          kustoClient,
+          KustoReadRequest(sparkSession, cachedSchema, kustoCoordinates, query, authentication, timeout, clientRequestProperties, requestId),
+          KustoFiltering(requiredColumns, filters))) match {
+          case Success(rdd) => rdd
```
ohadbitt, line 78: I think your indentation with 'Try match'  is confusing
please align the word 'case' with Try 
or new line the match and align it with Try

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
```diff
@@ -130,25 +128,28 @@ private[kusto] case class KustoRelation(kustoCoordinates: KustoCoordinates,
...
   private def getPartitioningColumn: String = {
-    if (readOptions.partitionOptions.column.isDefined) {
-      val requestedColumn = readOptions.partitionOptions.column.get
-      if (!schema.fields.exists(p => p.name equals requestedColumn)) {
-        throw new InvalidParameterException(
-          s"Cannot partition by column '$requestedColumn' since it is not part of the query schema: ${KDSU.NewLine}${schema.mkString(", ")}")
-      }
-      requestedColumn
-    } else schema.head.name
+    readOptions.partitionOptions.column match {
```
ohadbitt, line 131: Just FYI- partitioning is only on debug options and are not presented as user use case ..

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
```diff
@@ -38,79 +38,77 @@ private[kusto] case class KustoRelation(kustoCoordinates: KustoCoordinates,
...
+      // either a case of non-zero records or a case of timed-out.`
+      if(useSingleMode){
+        // there are less than 5000 (KustoConstants.DirectQueryUpperBoundRows) records, perform a single scan
+        Try(KustoReader.singleBuildScan(
+          kustoClient,
+          KustoReadRequest(sparkSession, cachedSchema, kustoCoordinates, query, authentication, timeout, clientRequestProperties, requestId),
+          KustoFiltering(requiredColumns, filters))) match {
+          case Success(rdd) => rdd
+          case Failure(exception) =>
+            KDSU.logError("KustoRelation","Failed with Single mode, falling back to Distributed mode," +
```
ohadbitt, line 80: Move log to the else clause

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
```diff
@@ -38,79 +38,77 @@ private[kusto] case class KustoRelation(kustoCoordinates: KustoCoordinates,
...
+              s"requestId: $requestId. Exception : ${exception.getMessage}")
+            // If the user specified forceSingleMode explicitly and that cannot be honored , throw an exception back
+            if(isUserOptionForceSingleMode){
+              // Expected behavior for Issue#261
+              throw exception
+            } else {
+              // The case where used did not provide an option and we estimated to be a single scan.
```
ohadbitt, line 87: add 
TODO: Parse exception. Fallback only if failed on 'too many rows'.

ohadbitt (changes requested):

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
ag-ramachandran: Added this parsing logic , In case that the exception on ForceSingleScan is too many rows, we fallback to distributed mode. All other exceptions for ForceSingleMode get thrown back and do not fallback

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
ag-ramachandran: Split this into readable lines

ag-ramachandran:
> lets add to the documentation that we suggest to provide a value here as estimation can take up to 6 seconds - or is too much details for a user?

I will add it as a note in the readme as ell

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoRelation.scala
```diff
@@ -38,117 +39,135 @@ private[kusto] case class KustoRelation(kustoCoordinates: KustoCoordinates,
...
+            // Only check is if exception is because of QueryLimits , it will fallback
+            val isRowLimitHit = ExceptionUtils.getRootCauseStackTrace(exception).contains("Query execution has exceeded the allowed limits")
+            if (isUserOptionForceSingleMode && !isRowLimitHit) {
+              // Expected behavior for Issue#261
+              throw exception
+            } else {
+              // The case where used did not provide an option and we estimated to be a single scan.
+              // Our approximate estimate failed here , fallback to distributed
+              KDSU.logError("KustoRelation", "Failed with Single mode, falling back to Distributed mode," +
+                s"requestId: $requestId. Exception : ${exception.getMessage}")
```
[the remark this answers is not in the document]
ohadbitt: You can also put the throw condition in the shouldNotThrow param

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoDataSourceUtils.scala
```diff
@@ -28,7 +28,7 @@ import java.net.URI
 import java.security.InvalidParameterException
 import java.util
 import java.util.concurrent.{Callable, CountDownLatch, TimeUnit}
-import java.util.{NoSuchElementException, Properties, StringJoiner, Timer, TimerTask, UUID}
+import java.util.{NoSuchElementException, Objects, Properties, StringJoiner, Timer, TimerTask, UUID}
```
ohadbitt, line 31: not used

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoDataSourceUtils.scala
```diff
@@ -642,25 +642,29 @@ object KustoDataSourceUtils {
...
-    val estimated = estimationResult.get(1)
-    if (estimated == null || StringUtils.isBlank(estimated.toString)) {
-      // Estimation can be empty for certain cases
+    val maybeEstimatedCount = Option(estimationResult.get(1))
+    /*
+     Check if the result is null or an empty string return a 0 , else return the numeric value
+    */
+    val estimatedCount = maybeEstimatedCount match {
+      case Some(ecStr: String) => if (StringUtils.isBlank(ecStr) || !StringUtils.isNumeric(ecStr)) /* Empty estimate */ 0 else ecStr.toInt
+      case Some(ecInt: java.lang.Number) => ecInt.intValue() // Is a numeric , get the int value back
```
ohadbitt, line 656: Did you check this logic ? im unfamiliar with java Numer ... Why not int?

ohadbitt (approved):

PR #264, opened 2022-10-06
new hidden options (so users do not use it without better knowledge) …

ohadbitt:
…and are under investigation. KUSTO_ENSURE_NO_DUPLICATED_BLOBS

- use tags to ingest blobs individually and ingest only if tag does not exist, meaning blob was not already ingested KUSTO_DISABLE_FLUSH_IMMEDIATELY - dont use flush immediately if clientBatchLimit reached

Bug fix : 
* Exception type for retrying of commands was wrong (was hiding real error) 
* Exceptions in queuing were not logged


ag-ramachandran:
Looks good to me , will just wait for user confirmation if these options work in the private build

PR #270, opened 2022-12-01
Bugfix/fix retry semantics

ohadbitt:
Fix bug on second retry (used same blob uuid)
Increase timeout to storage 
Move extents retry logic is changed to overlook permanent errors, increased retries after first batch success (10->1000).

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/utils/ExtendedKustoClient.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/utils/ExtendedKustoClient.scala
@@ -205,6 +205,8 @@ class ExtendedKustoClient(val engineKcsb: ConnectionStringBuilder, val ingestKcs
     var delayPeriodBetweenCalls = DelayPeriodBetweenCalls
     var consecutiveSuccesses = 0
     val useMaterializedViewFlag = shouldUseMaterializedViewFlag(database, targetTable, crp)
+    var firstMoveRetries = writeOptions.maxRetriesOnMoveExtents
+    var secondMovesRetries = 1000
     while (extentsProcessed < totalAmount) {
       var error: Object = null
       var res: Option[KustoResultSetTable] = None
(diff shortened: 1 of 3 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/utils/ExtendedKustoClient.scala
```diff
@@ -205,6 +205,8 @@ class ExtendedKustoClient(val engineKcsb: ConnectionStringBuilder, val ingestKcs
     var delayPeriodBetweenCalls = DelayPeriodBetweenCalls
     var consecutiveSuccesses = 0
     val useMaterializedViewFlag = shouldUseMaterializedViewFlag(database, targetTable, crp)
+    var firstMoveRetries = writeOptions.maxRetriesOnMoveExtents
```
ag-ramachandran, line 208: Nits, this and the one below can be vals. 
Yifat wanted the second threshold to be 10

connector/src/main/scala/com/microsoft/kusto/spark/utils/ExtendedKustoClient.scala
```diff
@@ -230,23 +232,24 @@ class ExtendedKustoClient(val engineKcsb: ConnectionStringBuilder, val ingestKcs
...
+//          }
         case ex: KustoDataExceptionBase =>
-          if (ex.getCause.isInstanceOf[SocketTimeoutException] || !ex.isPermanent) {
+//          if (ex.getCause.isInstanceOf[SocketTimeoutException] || !ex.isPermanent) {
             error = ExceptionUtils.getStackTrace(ex)
             failed = true
-          } else {
-            throw ex
-          }
+//          } else {
```
ag-ramachandran, line 250: may be we can remove this commented code

ag-ramachandran (approved):

PR #274, opened 2022-12-12
bug fix: Kusto real - cast int to double

ohadbitt:


connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoResponseDeserializer.scala
```diff
@@ -43,7 +43,11 @@ class KustoResponseDeserializer(val kustoResult: KustoResultSetTable) {
       case "int" => value: Any => value
       case "int32" => value: Any => value
       case "bool" => value: Any => value
-      case "real" => value: Any => value
+      case "real" => {
+        case v: Int => v.toDouble
```
ag-ramachandran, line 47: hello @ohadbitt since in each case we are doing toDouble, we can just do it for all cases instead of having a case match right case       `case "real" => value: Any => value.toString.toDouble` this may work and make it simple in all cases (we need a null check alone). What do you think

ag-ramachandran (approved):

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoResponseDeserializer.scala
ohadbitt: as discussed i dont want to call toString on each one as this is for every real value in the dataset

Issue #276, opened 2022-12-14
Add additional options for Export to DataStorage


ag-ramachandran:
Export to data storage has undergone enhancements that we want to use in the Spark framework. As an initial step, we want to be able to provide export options that users can configure in options. Since the options are not finite at this point, we can make this a property bag initially so that options can be expanded. 

For example: 

`
option("exportOptions"->"{"includeHeaders":"none"} ->-> results in -> .export .... with(includeHeaders=none} 
`

More on export: https://learn.microsoft.com/en-us/azure/data-explorer/kusto/management/data-export/export-data-to-storage

PR #278, opened 2022-12-21
Expose Export options for distributed mode

ag-ramachandran:
#### Pull Request Description

Add support for passing in additional options when using the **.export** model in ForceDistributedMode. Since there are lot of new options that can be passed in this is a map of options, passed in as a JSON

for example 
**'{"exportOption1":"e1", "exportOptionForCompression":"none"}'** 

When the **.export** is called in the ForceDistributedMode or because of limits getting hit these options will be added in the **with** clause

**.export async ..... with ( ..... , exportOption1="e1",exportOptionForCompression="none")**

Idea would be to take this as a KustoSourceOption and then validate it (has to be a proper name value pair as a parsable JSON)
and then use the map to provide options in export

Tests are WIP

**Features:**
- #276

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
@@ -4,7 +4,10 @@ import java.time.Instant
 import com.microsoft.kusto.spark.datasource.{TransientStorageCredentials, TransientStorageParameters}
 
 private[kusto] object CslCommandsGenerator {
-  private final val defaultKeySet = Set("compressionType","namePrefix","sizeLimit")
+  // We want to process these export options on a case to case basis. We want a default of snappy for compression
+  // and also want to ignore namePrefix because the files get exported with this name and then we need to read
+  // them with the same name in downstream processing.
+  private final val defaultKeySet = Set("compressionType","namePrefix","sizeLimit","compressed","async")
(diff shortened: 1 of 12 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
```diff
@@ -139,10 +141,13 @@ private[kusto] object CslCommandsGenerator {
...
     val sizeLimitIfDefined = if (sizeLimit.isDefined) s"sizeLimit=${sizeLimit.get * 1024 * 1024}, " else ""
-
+    val additionalOptionsString = additionalExportOptions.map {
+      case (k,v) => s"""$k="$v""""
+    }.mkString(",")
+    // TODO compressionType can be configured too and passed as an option
     var command =
       s""".export $async${compress}to parquet ("${storageParameters.storageCredentials.map(getFullUrlFromParams).reduce((s, s1) => s + ",\"" + s1)})""" +
-        s""" with (${sizeLimitIfDefined}namePrefix="${directory}part$partitionId", compressionType=snappy) <| $query"""
+        s""" with (${sizeLimitIfDefined}namePrefix="${directory}part$partitionId", compressionType=snappy,$additionalOptionsString) <| $query"""
```
ohadbitt, line 150: Will this work if additionalOptionsString has compressionType, part, namePrefix or sizeLimit?

ohadbitt (commented):
Looking good

asaharn (approved):

PR #280, opened 2023-01-12
Drop tags retry

ohadbitt:
Better retry for drop tags and async operations (could use it later in extents move .. or any async operation that permanent failure is considered transient)


ag-ramachandran (approved):
Correct me here, doing an E2E or a unit test here would be hard is'nt it. Do we want to add a TODO if want to add a test when we get to this completely

ag-ramachandran:
Hello @ohadbitt  Correct me here, doing an E2E or a unit test here would be hard is'nt it. Do we want to add a TODO if want to add a test when we get to this completely

ag-ramachandran (approved):

PR #281, opened 2023-01-16
Bug fix: distributed read could fail with Unsupported encoding: DELTA_BYTE_ARRAY 

ohadbitt:
Bug fix:
Disable new parquet writer on distributed reading if spark version is smaller than 3.3.0

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoReader.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoReader.scala
@@ -6,7 +6,8 @@ import com.microsoft.azure.storage.blob.CloudBlobContainer
 import com.microsoft.kusto.spark.authentication.KustoAuthentication
 import com.microsoft.kusto.spark.common.KustoCoordinates
 import com.microsoft.kusto.spark.datasource.ReadMode.ReadMode
-import com.microsoft.kusto.spark.utils.{CslCommandsGenerator, KustoAzureFsSetupCache, KustoBlobStorageUtils, ExtendedKustoClient, KustoDataSourceUtils => KDSU}
+import com.microsoft.kusto.spark.utils.{CslCommandsGenerator, ExtendedKustoClient, KustoAzureFsSetupCache, KustoBlobStorageUtils, KustoDataSourceUtils => KDSU}
+import org.apache.hadoop.util.ComparableVersion
 import org.apache.spark.Partition
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.sources.Filter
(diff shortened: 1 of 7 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoReader.scala
```diff
@@ -50,7 +51,7 @@ private[kusto] object KustoReader {
   private val myName = this.getClass.getSimpleName
   private val distributedReadModeTransientCache: concurrent.Map[DistributedReadModeTransientCacheKey,Seq[String]] =
     new concurrent.TrieMap()
-
+  private val minimalParquetWriterVersion = "3.3.0"
```
ag-ramachandran, line 54: May be we can add the following comment
/*
A new native implementation of Parquet writer that uses new encoding schemes was rolled out on the ADX side. This uses delta byte array for strings and other byte array-based Parquet types (default in Parquet V2 which most modern parquet readers support by default).
To avoid breaking changes for applications, if the runtime is on a lower version than 3.3.0 of spark runtime we explicitly set the ADX export to not use the useNativeIngestion
*/

connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
```diff
@@ -148,10 +149,12 @@ private[kusto] object CslCommandsGenerator {
     val compressionFormat = additionalExportOptions.getOrElse("compressionType", "snappy")
     val namePrefix = s"${directory}part$partitionId"
     val sizeLimitOverride = additionalExportOptions.get("sizeLimit").map(size => s"sizeLimit=${size.toLong * 1024 * 1024} ,").getOrElse("")
+    val nativeParquetString = additionalExportOptions.get("useNativeParquetWriter")
+      .map(b => s"useNativeParquetWriter=$b, ").getOrElse(if (!supportNewParquetWriter) "useNativeParquetWriter=false, " else "")
 
     var command =
       s""".export async $compress to parquet ("${storageParameters.storageCredentials.map(getFullUrlFromParams).reduce((s, s1) => s + ",\"" + s1)})""" +
-        s""" with ($sizeLimitOverride namePrefix="$namePrefix", compressionType="$compressionFormat"$additionalOptionsString) <| $query"""
```
ag-ramachandran, line 154: Quick question: Should we add in the Source options that if we use the useNativeParquetWriter and use 2.x versions of spark then , the additional options of VectorizedReader needs to be set (or) the exports will fail with the DELTA_BYTE_ARRAY error

connector/src/main/scala/com/microsoft/kusto/spark/datasource/KustoReader.scala
```diff
@@ -50,7 +51,7 @@ private[kusto] object KustoReader {
...
   private[kusto] def singleBuildScan(kustoClient: ExtendedKustoClient,
```
ag-ramachandran, line 55: Also, if okay we can add a TODO to add tests so that we will get to it later (while addressing that issue)

ag-ramachandran (approved):
Appoved the changes Ohad, left some minor comments more on documentation / code comments

PR #290, opened 2023-02-10
Kusto java sdk Version bump to 4.0.x

asaharn:
#### Pull Request Description

_Update Kusto java sdk to 4.0.x_

---

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- None


README.md
```diff
@@ -32,14 +32,14 @@ link your application with the artifact below to use the Azure Data Explorer Con
 ```
 groupId = com.microsoft.azure.kusto
 artifactId = kusto-spark_3.0_2.12
-version = 3.1.11
+version = 3.1.12
```
ag-ramachandran, line 35: The version should go to 4.0.x series

README.md
```diff
@@ -49,15 +49,15 @@ The jar can also be found under the [released package](https://github.com/Azure/
...
-        <version>3.1.11</version>
+        <version>3.1.12</version>
     </dependency>
 ```
 
 **In SBT**:
 
 ```scala
 libraryDependencies ++= Seq(
-  "com.microsoft.azure.kusto" %% "kusto-spark_3.0" % "3.1.11"
```
ag-ramachandran, line 60: Here too

connector/pom.xml
```diff
@@ -184,11 +196,6 @@
                 </exclusion>
             </exclusions>
         </dependency>
-        <dependency>
-            <groupId>org.json</groupId>
-            <artifactId>json</artifactId>
-            <version>${json.version}</version>
-        </dependency>
         <dependency>
```
ag-ramachandran, line 199: Java8 has all the API's that JODA has as well. Please create an issue to remove Joda time and we can just use stock Java8/11 datetime API

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -231,21 +234,25 @@ object KustoWriter {
...
+    }'$batchIdForTracing")
     val ingestClient = KustoClientCache.getClient(parameters.coordinates.clusterUrl,
       parameters.authentication, parameters.coordinates.ingestionUrl, parameters.coordinates.clusterAlias).ingestClient
     val queueRequestOptions = new QueueRequestOptions
     queueRequestOptions.setMaximumExecutionTimeInMs(KCONST.DefaultExecutionQueueing)
     queueRequestOptions.setTimeoutIntervalInMs(KCONST.DefaultTimeoutQueueing)
     queueRequestOptions.setRetryPolicyFactory(new RetryNoRetry)
-    ingestClient.setQueueRequestOptions(queueRequestOptions)
+    //TODO:Values need to be verified
+    val reqRetryOpts = new RequestRetryOptions(RetryPolicyType.EXPONENTIAL, Integer(KCONST.MaxIngestRetryAttempts), Duration.ofSeconds(KCONST.DefaultMaximumIngestionTime.toSeconds), null, null, null)
```
ag-ramachandran, line 251: Either verify the TODO using a test or remove it if it is fixed

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoDataSourceUtils.scala
```diff
@@ -160,13 +160,12 @@ object KustoDataSourceUtils {
...
-  private[kusto] def extractSchemaFromResultTable(result: Iterable[JSONObject]): String = {
+  private[kusto] def extractSchemaFromResultTable(result: ArrayNode): String = {
 
     val tableSchemaBuilder = new StringJoiner(",")
-
-    for (row <- result) {
+    for (i <- 0 until result.size()) {
       // Each row contains {Name, CslType, Type}, converted to (Name:CslType) pairs
-      tableSchemaBuilder.add(s"['${row.getString("Name")}']:${row.getString("CslType")}")
+      tableSchemaBuilder.add(s"['${result.get(i).get("Name").asText()}']:${result.get(i).get("CslType").asText()}")
```
ag-ramachandran, line 168: These constants Name and CSLType are used in few places. Would it be okay to refactor this into a SchemaConstants file and refer it from there?

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoIngestionUtils.scala
```diff
@@ -38,12 +37,12 @@ object KustoIngestionUtils {
...
 
   private[kusto] def setCsvMapping(sourceSchema: StructType,
-                                   targetSchema: Array[JSONObject],
+                                   targetSchema: Array[JsonNode],
                                    ingestionProperties: SparkIngestionProperties,tableCreationMode: SinkTableCreationMode): Unit = {
     require(ingestionProperties.csvMappingNameReference == null || ingestionProperties.csvMappingNameReference.isEmpty,
       "Sink options SparkIngestionProperties.csvMappingNameReference and adjustSchema.GenerateDynamicCsvMapping are not compatible. Use only one.")
 
-    val targetSchemaColumns = targetSchema.map(c => (c.getString("Name"),c.getString("CslType"))).toMap
+    val targetSchemaColumns = targetSchema.map(c => (c.get("Name").asText(),c.get("CslType").asText())).toMap
```
ag-ramachandran, line 45: Same as comment above, the constant strings

pom.xml
```diff
@@ -17,15 +17,15 @@
...
         <specs2.version>3.6.5</specs2.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven-source-plugin.version>3.0.1</maven-source-plugin.version>
-        <azure-storage.version>8.6.5</azure-storage.version>
+        <azure-storage.version>8.6.6</azure-storage.version>
         <keyvault-secrets.version>4.1.5</keyvault-secrets.version>
         <azure-identity.version>1.2.4</azure-identity.version>
         <commons-lang3.version>3.9</commons-lang3.version>
-        <fasterxml.jackson.version>[2.11.0,2.12.0)</fasterxml.jackson.version>
+        <fasterxml.jackson.version>[2.11.0,2.13.0)</fasterxml.jackson.version>
```
ag-ramachandran, line 28: Please do check this once with Databricks Jackson for library compatibility.

pom.xml
```diff
@@ -17,15 +17,15 @@
         <spark.version.minor>1</spark.version.minor>
 
         <!-- other dependencies -->
-        <kusto-sdk.version>3.2.1</kusto-sdk.version>
+        <kusto-sdk.version>4.0.3</kusto-sdk.version>
         <specs2.version>3.6.5</specs2.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven-source-plugin.version>3.0.1</maven-source-plugin.version>
-        <azure-storage.version>8.6.5</azure-storage.version>
+        <azure-storage.version>8.6.6</azure-storage.version>
```
ag-ramachandran, line 24: Is this storage consistent with the storage version we use in Java SDK ?

pom.xml
```diff
@@ -8,7 +8,7 @@
     <packaging>pom</packaging>
     <version>${revision}</version>
     <properties>
-        <revision>3.1.11</revision>
+        <revision>3.1.12</revision>
```
ag-ramachandran, line 11: Since this is a major change , may be 4.0.0 ?

pom.xml
```diff
@@ -17,15 +17,15 @@
...
-        <azure-storage.version>8.6.5</azure-storage.version>
+        <azure-storage.version>8.6.6</azure-storage.version>
         <keyvault-secrets.version>4.1.5</keyvault-secrets.version>
         <azure-identity.version>1.2.4</azure-identity.version>
         <commons-lang3.version>3.9</commons-lang3.version>
-        <fasterxml.jackson.version>[2.11.0,2.12.0)</fasterxml.jackson.version>
+        <fasterxml.jackson.version>[2.11.0,2.13.0)</fasterxml.jackson.version>
         <slf4j.version>1.8.0-beta4</slf4j.version>
         <joda-time.version>2.10.2</joda-time.version>
         <json.version>20201115</json.version>
```
ag-ramachandran, line 31: this lib json.version can go, no longer used

pom.xml
asaharn: They using 2.13.4 
@ag-ramachandran

pom.xml
asaharn: Seems this is now moved. And the same is used in Java SDK. We can also move from [com.azure](https://mvnrepository.com/artifact/com.azure) » azure-storage-blob
image

asaharn: Will open a new issue for this.

ag-ramachandran: sure, would it be possible to see what changes will be needed for this. May be try the update in this same PR ?

asaharn: Sure. There are some new classes introduced, I guess. Will check them out.

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoIngestionUtils.scala
asaharn: Done

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -231,21 +234,20 @@ object KustoWriter {
...
+        requestId
+    }'$batchIdForTracing")
     val ingestClient = KustoClientCache.getClient(parameters.coordinates.clusterUrl,
       parameters.authentication, parameters.coordinates.ingestionUrl, parameters.coordinates.clusterAlias).ingestClient
-    val queueRequestOptions = new QueueRequestOptions
-    queueRequestOptions.setMaximumExecutionTimeInMs(KCONST.DefaultExecutionQueueing)
-    queueRequestOptions.setTimeoutIntervalInMs(KCONST.DefaultTimeoutQueueing)
-    queueRequestOptions.setRetryPolicyFactory(new RetryNoRetry)
-    ingestClient.setQueueRequestOptions(queueRequestOptions)
+    val reqRetryOpts = new RequestRetryOptions(RetryPolicyType.EXPONENTIAL, Integer(KCONST.MaxIngestRetryAttempts), Duration.ofSeconds(KCONST.DefaultMaximumIngestionTime.toSeconds), null, null, null)
```
ohadbitt, line 246: This is very different then what was before!
The reason we had NoRetry here is that we favor longer timeout over retries when it comes to queueing or else we get duplications

pom.xml
ohadbitt: Actually if this is the PR where we upgrade major version this has to be on this PR
java sdk removed all dependencies on the old sdk

pom.xml
ohadbitt: then lets use 2.13.4 as well

ohadbitt:
What's really important is that you check our read and write scenarios on ADB with different runtimes
I think it would only work in runtime > 12 .. 
make sure that your test invokes every flow that involves jackson library as what we are afraid of is that its going to fail on runtime looking for these.

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
ohadbitt: You can do something similar to .Net
https://msazure.visualstudio.com/One/_git/Azure-Kusto-Service?path=/Src/Client/Kusto.Ingest/Storage/QueueOptions.cs

connector/pom.xml
```diff
@@ -14,6 +14,18 @@
 
     <dependencies>
```
ohadbitt, line 15: To avoid version collisions, we moved in the sdk in this version to using azure bom
Please use azure bom instead of specifying version of every   <groupId>com.azure</groupId> 
dependency
https://github.com/Azure/azure-kusto-java/blob/master/data/pom.xml
https://github.com/Azure/azure-kusto-java/blob/master/ingest/pom.xml

`<dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>com.azure</groupId>
                <artifactId>azure-sdk-bom</artifactId>
                <version>${azure-bom-version}</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>`

connector/pom.xml
```diff
@@ -14,6 +14,18 @@
...
         <!-- Compile Scope -->
```
ohadbitt, line 16: Please update `                <artifactId>scala-maven-plugin</artifactId>`
To use   target:jvm-11 (or however its supposed to be called)

ohadbitt (changes requested):

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
ohadbitt: Not fixed

connector/pom.xml
```diff
@@ -108,7 +125,7 @@
         <dependency>
             <groupId>com.azure</groupId>
```
ohadbitt, line 126: This is also part of the bom management

connector/src/main/scala/com/microsoft/kusto/spark/datasink/FinalizeHelper.scala
```diff
@@ -142,10 +141,10 @@ object FinalizeHelper {
           finalRes = Some(partitionResult.ingestionResult.getIngestionStatusCollection.get(0))
           finalRes
         } catch {
-          case e: StorageException =>
-            KDSU.logWarn(loggerName, "Failed to fetch operation status transiently - will keep polling. " +
-              s"RequestId: $requestId. Error: ${ExceptionUtils.getStackTrace(e)}")
-            None
+//          case e: RequestFailedException =>
```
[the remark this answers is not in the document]
asaharn: Should we be doing something different operation with that? or is it just logging? I tried looking about this Exception but couldn't find much.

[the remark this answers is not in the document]
ohadbitt: Its an exception raised from request to table - so we need to catch only in that case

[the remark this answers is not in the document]
ohadbitt: We actually should not throw here - so change it to 
          case e: TableServiceErrorException => KDSU.reportExceptionAndThrow(loggerName, e, s"TableServiceErrorException : RequestId: $requestId", shouldNotThrow = true)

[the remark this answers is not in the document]
ohadbitt: maybe we want to count retries or something but this can be a different PR

connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoClientCache.scala
```diff
@@ -66,8 +67,9 @@ object KustoClientCache {
       )
     }
 
-    engineKcsb.setClientVersionForTracing(KCONST.ClientName)
-    ingestKcsb.setClientVersionForTracing(KCONST.ClientName)
+    val sparkVersion = SparkSession.builder().getOrCreate().version
+    engineKcsb.setConnectorDetails(KCONST.ClientName, KustoDataSourceUtils.Version ,null, null, false, null,  Pair.of("spark.version", sparkVersion))
```
ohadbitt, line 71: we can actually get the application name 
we have it in KustoWriter - maybe you can get it here the same way as well
data.sparkSession.sparkContext.appName

ohadbitt (approved):

PR #291, opened 2023-02-21
* Added support for .move extents time range

asaharn:
#### Pull Request Description

_Improved performance by providing timerange to .move extents_

---

#### Future Release Comment
_[Add description of your change, to include in the next release]_
_[Delete any or all irrelevant sections, e.g. if your change does not warrant a release comment at all]_

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- None

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
@@ -32,7 +32,7 @@ import java.util.zip.GZIPOutputStream
 import java.util.{TimeZone, UUID}
 import com.microsoft.kusto.spark.datasink.FinalizeHelper.finalizeIngestionWhenWorkersSucceeded
 
-import java.time.Instant
+import java.time.{Clock, Instant}
 import scala.collection.JavaConverters._
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.util.{Failure, Success, Try}
@@ -177,7 +177,7 @@ object KustoWriter {
     val creationTime = Option(ingestionProperties.toIngestionProperties(tableCoordinates.database, tableCoordinates.table.get).getAdditionalProperties.get("creationTime"))
     creationTime match {
       case Some(creationTimeVal) => Instant.parse(creationTimeVal)
-      case None => Instant.now()
+      case None => Instant.now(Clock.systemUTC())
     }
   }
 
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
@@ -174,8 +174,9 @@ object KustoWriter {
   }
 
   def getCreationTime(ingestionProperties: SparkIngestionProperties, tableCoordinates: KustoCoordinates): Instant = {
-    val creationTime = Option(ingestionProperties.toIngestionProperties(tableCoordinates.database, tableCoordinates.table.get).getAdditionalProperties.get("creationTime"))
-    creationTime match {
+    val startTime = Option(ingestionProperties.toIngestionProperties(tableCoordinates.database, tableCoordinates.table.get).getAdditionalProperties.get("startTime"))
+
+    startTime match {
       case Some(creationTimeVal) => Instant.parse(creationTimeVal)
       case None => Instant.now(Clock.systemUTC())
     }
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
@@ -128,6 +128,7 @@ object KustoWriter {
       val partitionsResults = rdd.sparkContext.collectionAccumulator[PartitionResult]
       val parameters = KustoWriteResource(authentication = authentication, coordinates = tableCoordinates,
         schema = data.schema, writeOptions = rebuiltOptions, tmpTableName = tmpTableName)
+      val sinkStartTime = Instant.now()
       if (writeOptions.isAsync) {
         val asyncWork = rdd.foreachPartitionAsync { rows => ingestRowsIntoTempTbl(rows, batchIdIfExists,
           partitionsResults,parameters) }
@@ -138,7 +139,7 @@ object KustoWriter {
           asyncWork.onSuccess {
             case _ => finalizeIngestionWhenWorkersSucceeded(
               tableCoordinates, batchIdIfExists, tmpTableName, partitionsResults,
-              writeOptions, crp, tableExists, rdd.sparkContext, authentication, kustoClient)
+              writeOptions, crp, tableExists, rdd.sparkContext, authentication, kustoClient, sinkStartTime)
           }
           asyncWork.onFailure {
             case exception: Exception =>
@@ -166,7 +167,7 @@ object KustoWriter {
         if (writeOptions.isTransactionalMode) {
           finalizeIngestionWhenWorkersSucceeded(
           tableCoordinates, batchIdIfExists, tmpTableName, partitionsResults, writeOptions,
-          crp, tableExists, rdd.sparkContext, authentication, kustoClient)
+          crp, tableExists, rdd.sparkContext, authentication, kustoClient, sinkStartTime)
         }
       }
     }
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
@@ -128,7 +128,7 @@ object KustoWriter {
       val partitionsResults = rdd.sparkContext.collectionAccumulator[PartitionResult]
       val parameters = KustoWriteResource(authentication = authentication, coordinates = tableCoordinates,
         schema = data.schema, writeOptions = rebuiltOptions, tmpTableName = tmpTableName)
-      val sinkStartTime = Instant.now()
+      val sinkStartTime = getCreationTime(stagingTableIngestionProperties, tableCoordinates)
       if (writeOptions.isAsync) {
         val asyncWork = rdd.foreachPartitionAsync { rows => ingestRowsIntoTempTbl(rows, batchIdIfExists,
           partitionsResults,parameters) }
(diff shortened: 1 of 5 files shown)


connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -128,6 +128,7 @@ object KustoWriter {
       val partitionsResults = rdd.sparkContext.collectionAccumulator[PartitionResult]
       val parameters = KustoWriteResource(authentication = authentication, coordinates = tableCoordinates,
         schema = data.schema, writeOptions = rebuiltOptions, tmpTableName = tmpTableName)
+      val sinkStartTime = Instant.now()
```
y0nil, line 131: please note that in some clients/pipelines, customers can override the `creationTime` of extents created by ingestion as part of ingestion properties, in which case - the minimum and maximum values for the datetime range should be based on the properties they pass.
if this isn't supported in the Spark connector - then all is good.
if it is, then this will need to be adjusted accordingly (i.e. if `creationTime` is specified - use that instead of `Instants.now()`)

asaharn: Done.

y0nil: i think you misunderstood. let me try to explain differently
- you don't have to change the creationTime property. only in rare cases do customers choose to do so.
- if _the customer_ hasn't chosen to set the 'creationTime' property, then using the current datetime (UTC now) to capture the job's start/end time is just fine.

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -166,12 +167,20 @@ object KustoWriter {
...
         }
       }
     }
   }
 
+  def getCreationTime(ingestionProperties: SparkIngestionProperties, tableCoordinates: KustoCoordinates): Instant = {
+    val creationTime = Option(ingestionProperties.toIngestionProperties(tableCoordinates.database, tableCoordinates.table.get).getAdditionalProperties.get("creationTime"))
+    creationTime match {
+      case Some(creationTimeVal) => Instant.parse(creationTimeVal)
+      case None => Instant.now()
```
ag-ramachandran, line 180: `Instant.now(Clock.systemUTC())` Is the best option to use to be sure.

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -166,12 +167,20 @@ object KustoWriter {
...
+          crp, tableExists, rdd.sparkContext, authentication, kustoClient, sinkStartTime)
         }
       }
     }
   }
 
+  def getCreationTime(ingestionProperties: SparkIngestionProperties, tableCoordinates: KustoCoordinates): Instant = {
+    val creationTime = Option(ingestionProperties.toIngestionProperties(tableCoordinates.database, tableCoordinates.table.get).getAdditionalProperties.get("creationTime"))
+    creationTime match {
+      case Some(creationTimeVal) => Instant.parse(creationTimeVal)
```
ag-ramachandran, line 179: Do we validate this date time format upfront ? If we already have it in the documentation its okay. We have to explicitly call out and even better if we validate this upfront when we parse the options.
If not now, we can add a TODO that we need to add this at some point

ag-ramachandran (changes requested):

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
ag-ramachandran: Please correct us here @y0nil . The logic was the following. If the user overrides the extentCreationTime, then we set it this way or else fallback to now. Did we miss something here.

```
  def getCreationTime(ingestionProperties: SparkIngestionProperties, tableCoordinates: KustoCoordinates): Instant = {
    val creationTime = Option(ingestionProperties.toIngestionProperties(tableCoordinates.database, tableCoordinates.table.get).getAdditionalProperties.get("creationTime"))
    creationTime match {
      case Some(creationTimeVal) => Instant.parse(creationTimeVal)
      case None => Instant.now()
```

y0nil: no, you don't need to explicitly set creationTime. only if the end-user specified is should it be set.
otherwise, it shouldn't be explicitly set.

asaharn: Yes. Agreed Yoni. I guess that's what we did here? We are not setting the creationTime anywhere, but just taking the creationTime (only if exists) in first place, else the current time.

y0nil: let me try to explain again - you should never set creationTime to "Instant.now()". you should only set it if it was provided by the end user

asaharn: I think you are confusing the local variable creationTime with the ingestion property `creationTime`. Sorry for the confusing name will change it. 
```
def getCreationTime(ingestionProperties: SparkIngestionProperties, tableCoordinates: KustoCoordinates): Instant = {
    val creationTime = Option(ingestionProperties.toIngestionProperties(tableCoordinates.database, tableCoordinates.table.get).getAdditionalProperties.get("creationTime"))
    creationTime match {
      case Some(creationTimeVal) => Instant.parse(creationTimeVal)
      case None => Instant.now()
```

The local variable, creationTime, will be set if the ingestion prop `creationTime` is provided by user. Else the current Instant will be set. and returned.

y0nil: yes, that's much clearer than the previous version you quoted

```
    creationTime match {
      case Some(creationTimeVal) => Instant.parse(creationTimeVal)
      case None => Instant.now()
```

ag-ramachandran: Thanks @y0nil, with the semantic change does it sound okay :) if so , will fold it into master and plan a release this week.

connector/src/main/scala/com/microsoft/kusto/spark/utils/CslCommandsGenerator.scala
```diff
@@ -81,22 +82,23 @@ private[kusto] object CslCommandsGenerator {
     s""".show table ${tableName} details | project todynamic(ShardingPolicy).UseShardEngine"""
   }
 
-  def generateTableMoveExtentsCommand(sourceTableName: String, destinationTableName: String, batchSize: Int,
+  def generateTableMoveExtentsCommand(sourceTableName: String, destinationTableName: String, timerange: Array[Instant], batchSize: Int,
                                       isDestinationTableMaterializedViewSource: Boolean = false): String = {
     val setNewIngestionTime: String = if (isDestinationTableMaterializedViewSource) "with(SetNewIngestionTime=true)" else ""
     s""".move extents to table $destinationTableName $setNewIngestionTime <|
-       .show table $sourceTableName extents with(extentsShowFilteringRuntimePolicy='{"MaximumResultsCount":$batchSize}');
+       .show table $sourceTableName extents with(extentsShowFilteringRuntimePolicy='{"MaximumResultsCount":$batchSize}') with(extentCreatedOnFrom='${timerange(0)}', extentCreatedOnTo='${timerange(1)}');
```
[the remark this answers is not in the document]
asaharn: Hi Yoni,

Yes we have E2E tests that checks the functionality by writing the data and reading that back again to check if both are equal. And the above code passed them. 

After your suggestion above i.e. `.move extents to table $destinationTableName with(setNewIngestionTime=<true of false>, extentCreatedOnFrom='...', extentCreatedOnTo='...', extentsShowFilteringRuntimePolicy='{"MaximumResultsCount":...}') <| .show table $sourceTableName extents`, 
got the below findings:

1) I took the final command built by code (in current scenario) and ran it manually which ran without any issues with the desired result. 
```kql
.move async extents to table KustoSparkReadWriteTest_f84ac15e_3f51_49cf_a6f0_bbb70915xxxx  <|
       .show table sparkTempTable_KustoSink_KustoSparkReadWriteTest_xxxxx_3f51_49cf_a6f0_bbb70915018f_3004f1b6_1599_49a8_91cc_xxxxx
       extents with(extentsShowFilteringRuntimePolicy='{"MaximumResultsCount":1}')
       with(extentCreatedOnFrom='2023-03-02T11:28:41.474Z', extentCreatedOnTo='2023-03-02T11:29:02.405063200Z');
```

2) After making the changes asked above, i.e. moving all the props in a single with we got exceptions:
```kql
.move async extents to table KustoSparkReadWriteTest_f84ac15e_3f51_49cf_a6f0_bbb70915xxxx with ( extentsShowFilteringRuntimePolicy='{"MaximumResultsCount":1}', extentCreatedOnFrom='2023-03-02T11:28:41.474Z', extentCreatedOnTo='2023-03-02T11:29:02.405063200Z') <|
       .show table sparkTempTable_KustoSink_KustoSparkReadWriteTest_xxxxx_3f51_49cf_a6f0_bbb70915018f_3004f1b6_1599_49a8_91cc_xxxxx
       extents
```
 got the below error:
image


Which make sense to me, since we cannot move prop `extentsShowFilteringRuntimePolicy` before `.show table extents` command. 

Really appreciate your suggestions, or to wrap this faster we can have a 15min call?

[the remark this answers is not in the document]
y0nil: sorry for the confusion.

instead of this:

```
.move async extents to table KustoSparkReadWriteTest_f84ac15e_3f51_49cf_a6f0_bbb70915xxxx with ( extentsShowFilteringRuntimePolicy='{"MaximumResultsCount":1}', extentCreatedOnFrom='2023-03-02T11:28:41.474Z', extentCreatedOnTo='2023-03-02T11:29:02.405063200Z') <|
       .show table sparkTempTable_KustoSink_KustoSparkReadWriteTest_xxxxx_3f51_49cf_a6f0_bbb70915018f_3004f1b6_1599_49a8_91cc_xxxxx
       extents
```

it should be this:

```
.move async extents to table KustoSparkReadWriteTest_f84ac15e_3f51_49cf_a6f0_bbb70915xxxx with (extentCreatedOnFrom='2023-03-02T11:28:41.474Z', extentCreatedOnTo='2023-03-02T11:29:02.405063200Z') <|
       .show table sparkTempTable_KustoSink_KustoSparkReadWriteTest_xxxxx_3f51_49cf_a6f0_bbb70915018f_3004f1b6_1599_49a8_91cc_xxxxx
       extents with(extentsShowFilteringRuntimePolicy='{"MaximumResultsCount":1}')
```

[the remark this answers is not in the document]
asaharn: Thanks @y0nil. Have added the changes.

y0nil (approved):

ag-ramachandran (approved):

PR #297, opened 2023-03-17
* Added a ReadWrite sample for Trident

ag-ramachandran:
#### Pull Request Description

_Add a sample for Trident_

---

**Breaking Changes:**
- None

**Features:**
- Add a notebook working sample for Trident

**Fixes:**
- None


samples/src/main/trident/NYC-GreenTaxi-Read-Write-Data-To-Kusto.ipynb
```diff
@@ -0,0 +1 @@
...assengerCount","type":"int"},{"key":"4","name":"tripDistance","type":"double"},{"key":"5","name":"puLocationId","type":"string"},{"key":"6","name":"doLocationId","type":"string"},{"key":"7","name":"pickupLongitude","type":"double"},{"key":"8","name":"pickupLatitude","type":"double"},{"key":"9","name":"dropoffLongitude","type":"double"},{"key":"10","name":"dropoffLatitude","type":"double"},{"key":"11","name":"rateCodeID","type":"int"},{"key":"12","name":"storeAndFwdFlag","type":"string"},{"key":"13","name":"paymentType","type":"int"},{"key":"14","name":"fareAmount","type":"double"},{"key":"15","name":"extra","type":"double"},{"key":"16","name":"mtaTax","type":"double"},{"key":"17","name":"improvementSurcharge","type":"string"},{"key":"18","name":"tipAmount","type":"double"},{"key":"19","name":"tollsAmount","type":"double"},{"key":"20","name":"ehailFee","type":"double"},{"key":"21","name":"totalAmount","type":"double"},{"key":"22","name":"tripType","type":"int"},{"key":"23","name":"puYear","type":"int"},{"key":"24","name":"puMonth","type":"int"}],"truncated":false}},"type":"Synapse.DataFrame"}},"version":"0.1"},"trident":{"lakehouse":{}},"widgets":{}},"nbformat":4,"nbformat_minor":0}
```
ohadbitt, line 1: Why do we need this?
print('Register the DataFrame as a SQL temporary view: source')
df.createOrReplaceTempView('source')

ohadbitt, line 1: Change
 The targets where this data has to be written
to
 The target where this data will be written to

ohadbitt, line 1: Consider changing this:
The access token in the current case is the user's credential that is used to write the data to the GreenTaxiData table in Kusto.

The access token is created using the user's credential and will be used to write the data to the Kusto table GreenTaxiData, therefore the user is required for 'user' privileges or above on the target database
and table 'admin' privileges if the table already exists. If the table does not exist, it will be created with the DataFrame schema.

ohadbitt (changes requested):

ohadbitt (approved):

Issue #222, opened 2022-02-16
Support streaming ingest


epa095:
I am looking into how low latency I can get between spark and ADX. And I see that one way to get lower latency is to enable [Streaming ingestion policy](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/management/streamingingestionpolicy) in both the ADX cluster and database/table. But after enabling I still see that it takes 5 minutes for my batches from spark (through this connector) to arive. 

The [documentation](https://github.com/Azure/azure-kusto-spark/blob/52b6782be8abfc6eea028cc341f951914e19a98b/docs/KustoSink.md#performance-considerations) only mentiones [Ingestion Batching Policy](https://docs.microsoft.com/en-us/azure/kusto/concepts/batchingpolicy), making me think that maybe this connector does not support using the streaming ingestion into ADX? If so, it would be very nice (and natural) if it started supporting it. Maybe [this](https://github.com/Azure/azure-kusto-python/pull/344) new feature in [azure-kusto-python](https://github.com/Azure/azure-kusto-python) makes it easier?

yogilad:
Same issue as #217

yogilad:
@epa095, 
Spark uses temporary tables, which don't adhere to the target table ingestion batching policy configuration.
We have a bug on this (above).

As a work around, if this is ok for your specific case, you can set the ingestion batching policy on the database level.
You can also consider one of the following workarounds
•	Lowering KustoSink’s “KUSTO_CLIENT_BATCHING_LIMIT” config from its default of 100MB
. Threshold for Spark to ship its data to the Kusto service.
•	Setting the “KustoSinkOptions.KUSTO_SPARK_INGESTION_PROPERTIES_JSON.[flushImmediately] config to true
 on The Kusto service will ignore the ingestion batching policy I described earlier, and ingest immediately without batching.

epa095:
@yogilad : I have changed the policy on the *database* level, and both enabled streamingingestion and set ingestionbatching - MaximumBatchingTimeSpan to 10 seconds. By setting KustoSinkOptions.KUSTO_SPARK_INGESTION_PROPERTIES_JSON.[flushImmediately] to true I get the following timings from spark using this sink:
```json
  "durationMs" : {
    "addBatch" : 4826,
    "getBatch" : 32,
    "latestOffset" : 30,
    "queryPlanning" : 5,
    "triggerExecution" : 5093,
    "walCommit" : 111
  },
```
[...]

PR #301, opened 2023-04-12
Kusto Stream Ingestion

jrob5756:
#### Pull Request Description

This pull request contains code (& tests) to introduce [kusto streaming ingestion](https://learn.microsoft.com/en-us/azure/data-explorer/kusto/management/streamingingestionpolicy) support in the Spark Kusto Connector. 
Stream ingestion is enabled by setting the [KUSTO_WRITE_MODE](https://github.com/Azure/azure-kusto-spark/blob/master/docs/KustoSink.md#supported-options) option to `Stream`. 

Notes
-  Stream ingestion must be enabled on the destination Table and/or Cluster (see [here](https://learn.microsoft.com/en-us/azure/data-explorer/ingest-data-streaming?tabs=azure-portal%2Cjava) for details).
- Stream ingestion has a [data size limit](https://learn.microsoft.com/en-us/azure/data-explorer/kusto/management/streamingingestionpolicy) of 4 MB. If a partition exceeds this size, the connector will send multiple appropriately sized requests.

---

**Breaking Changes:**
- None

**Features:**
- https://github.com/Azure/azure-kusto-spark/issues/222

**Fixes:**
- None

### DIFF
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoSinkOptions.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoSinkOptions.scala
@@ -57,6 +57,8 @@ object KustoSinkOptions extends KustoOptions{
   // after which it will move the data to the destination table (the last part is a metadata operation only)
   // If set to 'Queued', the write operation finishes after data is processed by the workers, the data may not be completely
   // available up until the service finishes loading it and failures on the service side will not propagate to Spark.
+  // If set to 'Stream', Kusto streaming ingestion will be used. Streaming ingestion should be used if latency of less than a few seconds is required
+  // or To optimize operational processing of many tables where the stream of data into each table is relatively small (a few records per second).
   val KUSTO_WRITE_MODE: String = newOption("writeMode")
 
   // Provide a temporary table name that will be used for this write operation to achieve transactional write and move
@@ -82,7 +84,7 @@ object SchemaAdjustmentMode extends Enumeration {
 
 object WriteMode extends Enumeration {
   type WriteMode = Value
-  val Transactional, Queued = Value
+  val Transactional, Queued, Stream = Value
 }
 
 case class WriteOptions(pollingOnDriver:Boolean = false,
@@ -100,7 +102,7 @@ case class WriteOptions(pollingOnDriver:Boolean = false,
                         maxRetriesOnMoveExtents: Int = 10,
                         minimalExtentsCountForSplitMerge: Int = 400,
                         adjustSchema: SchemaAdjustmentMode.SchemaAdjustmentMode = SchemaAdjustmentMode.NoAdjustment,
-                        isTransactionalMode: Boolean = true,
+                        writeMode: WriteMode.WriteMode = WriteMode.Transactional,
                         userTempTableName: Option[String] = None,
                         disableFlushImmediately:Boolean = false,
                         ensureNoDupBlobs: Boolean = false)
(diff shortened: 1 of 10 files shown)


ag-ramachandran:
Hello @jrob5756 , Thanks for the PR. 

We're going to have a look at this and provide a review on this in a week or 2 ( we are in the process of a version bump for JDK 11 and up).

ohadbitt:
We need to discuss internally the way we want to deal with streaming, 
From the PR note : Stream ingestion has a [data size limit](https://learn.microsoft.com/en-us/azure/data-explorer/kusto/management/streamingingestionpolicy) of 4 MB. If a micro-batch exceeds this size, the connector will fall back to the Queued write mode.
For clarification, the decision made here is per partition and not per micro-batch, which is ok- jsut calling it by the right name..
Anyway I think we should try and see if its possible to split the data into 4 MB chunks instead and try to stream each one individually. There are cons to this as well - as we might be trying to stream this way GBs of data which is not ideal - maybe a big partition does indicate its not streaming material.
Another issue is that we would probably want the ManagedStreaming client instead of plain streaming.. cons of that is that managed streaming doesn't provide Transactionality (same with the 4mb Queued fallback ... ) .. so we need to discuss this further as well because Streaming is Transactional by definition.

jrob5756:
> We need to discuss internally the way we want to deal with streaming, From the PR note : Stream ingestion has a [data size limit](https://learn.microsoft.com/en-us/azure/data-explorer/kusto/management/streamingingestionpolicy) of 4 MB. If a micro-batch exceeds this size, the connector will fall back to the Queued write mode. For clarification, the decision made here is per partition and not per micro-batch, which is ok- jsut calling it by the right name.. Anyway I think we should try and see if its possible to split the data into 4 MB chunks instead and try to stream each one individually. There are cons to this as well - as we might be trying to stream this way GBs of data which is not ideal - maybe a big partition does indicate its not streaming material. Another issue is that we would probably want the ManagedStreaming client instead of plain streaming.. cons of that is that managed streaming doesn't provide Transactionality (same with the 4mb Queued fallback ... ) .. so we need to discuss this further as well because Streaming is Transactional by definition. @ag-ramachandran - lets discuss this internally and then resort to reviewing this PR

Thanks @ohadbitt. Good call on the microbatch vs. partition comment. I updated the pr summary to reflect that. Regarding the Queued fallback approach, I had the same thoughts while I was implementing these changes. As you pointed out, there are pros and cons to both strategies. If the team is more comfortable splitting the data into 4 MB chunks, I can certainly do that as it would not be difficult to implement. Let me know!

jrob5756:
@ag-ramachandran quick update. We discussed offline but, adding comments here as well for transparency. Since the PR was created the code has been updated to chunk the requests as discussed above. Also, I saw that there have been some structural changes recently which were incompatible with the PR. I've updated the code so we should be squared away there as well. I await your review...

Raghu0210:
@jrob5756 @ag-ramachandran
As a user, I truly appreciate the amount of work you are putting into this feature. Any update on the **timeline** for this, much awaited feature (spark-kusto streaming ingest).

We are working on a POC to stream large volume of data using Spark-Kusto connector (Databricks) into ADX streaming tables ingesting with a requirement of very low latency(ms).

Thanks for your time and dedication for developing this feature. Looking forward to your update.

ag-ramachandran:
> @jrob5756 @ag-ramachandran As a user, I truly appreciate the amount of work you are putting into this feature. Any update on the **timeline** for this, much awaited feature (spark-kusto streaming ingest).
> 
> We are working on a POC to stream large volume of data using Spark-Kusto connector (Databricks) into ADX streaming tables ingesting with a requirement of very low latency(ms).
> 
> Thanks for your time and dedication for developing this feature. Looking forward to your update.

Hello @Raghu0210 , we should get this picked up by the end of this year. The challenge has been in changes we have been making in the connector and this taking a lower priority. Will keep this thread updated

PR #318, opened 2023-06-14
* Fix the bug with exception not propagated

ag-ramachandran:
#### Pull Request Description

Fix the issue where the exception is not propagated in case of Queued Mode

---

#### Future Release Comment
Will need to be cherrypicked for 2.4 as well as 4.x versions


asaharn (approved):

ohadbitt (approved):
make sure you also create a version for master

PR #326, opened 2023-07-05
added Resource Manager as per new Kusto sdk

asaharn:
#### Pull Request Description

_[Add a description of your pull request here]_

---

#### Future Release Comment
_[Add description of your change, to include in the next release]_
_[Delete any or all irrelevant sections, e.g. if your change does not warrant a release comment at all]_

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- None

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,11 +17,13 @@
         <!-- other dependencies -->
         <azure.bom.version>1.2.15</azure.bom.version>
         <!-- versions from https://learn.microsoft.com/en-us/azure/databricks/release-notes/runtime/13.2-->
-        <fasterxml.jackson.jsr.version>2.13.4</fasterxml.jackson.jsr.version>
+        <fasterxml.jackson.jar.version>2.13.4</fasterxml.jackson.jar.version>
         <fasterxml.jackson.version>2.14.3</fasterxml.jackson.version>
         <azure.bom.version>1.2.4</azure.bom.version>
         <!-- Deprecate only after full migration to azure storage 12-->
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -21,7 +21,7 @@
         <fasterxml.jackson.databind.version>2.13.4.2</fasterxml.jackson.databind.version>
         <jsonsmart.version>2.4.10</jsonsmart.version>
         <hadoop.version>3.3.5</hadoop.version>
-        <kusto.sdk.version>4.0.4</kusto.sdk.version>
+        <kusto.sdk.version>5.0.0</kusto.sdk.version>
```
ohadbitt, line 24: change when released

ohadbitt (commented):
What about the export containers ? 
Can we somehow use this to do the same?

ag-ramachandran:
> What about the export containers ? Can we somehow use this to do the same?

Hello @ohadbitt , Great question :) . 

We had a quick discussion and got inputs on this with @yogilad. They are similar and dissimilar in a few ways

Focusing on dissimilarity:

a) The major difference is that we get the storage and SAS and we can make a note of the bad storage accounts and use it lower on the queue when get resources is invoked

b) In case of export containers, the data is written to the storage from the engine and though the export may fail, it will need work to understand if this was a transient network fail vs a permanent storage failure and rank the storage

The way to do it would be read storage metadata when we get the SAS (so that we know the storage/SAS is good) and that will need more work from the client. We can get to it (defer for now)

What are your thoughts

PR #347, opened 2023-11-21
* Retries for DM not reachable

ag-ramachandran:
#### Pull Request Description

We had the error which was reported by the ADF folks for Host not being reachable. Do you think this makes sense to retry that request


connector/src/test/scala/com/microsoft/kusto/spark/utils/ContainerProviderTest.scala
```diff
@@ -75,6 +80,57 @@ class ContainerProviderTest extends AnyFlatSpec with Matchers with MockFactory {
...
+    val ingestProviderEntryCreator = (c: ContainerAndSas) => c
+    /*
+      Invoke and test
+     */
+    val containerProvider = new ContainerProvider(extendedMockClient, clusterAlias, command,
+      ingestProviderEntryCreator, CACHE_EXPIRY_SEC)
+
+    extendedMockClient.executeDM _ expects(command, None, *) throws new DataServiceException(clusterAlias,
+      "IOError when trying to retrieve CloudInfo",(new HttpHostConnectException(
+      new IOException("Connection timed out"),
```
ohadbitt, line 99: I think the error is built 
new HttpHostConnectException(new java.net.ConnectException("..

ohadbitt (approved):

PR #363, opened 2024-02-01
* Fix scala format

ag-ramachandran:
The only changes are in the pom.xml file. The others are just format runs

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,0 +1,34 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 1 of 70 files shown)


.scalafmt.conf
```diff
@@ -0,0 +1,34 @@
...
+align = none
+align.openParenDefnSite = false
+align.openParenCallSite = false
+align.tokens = []
+optIn = {
+  configStyleArguments = false
+}
+danglingParentheses.preset = false
+docstrings.style = Asterisk
+maxColumn = 98
```
ohadbitt, line 27: 98 is row length ? 120 or 160 are other langs standards

connector/pom.xml
```diff
@@ -357,6 +357,16 @@
                 </executions>
                 <configuration>
                     <addScalacArgs>-target:jvm-1.8</addScalacArgs>
+                    <args>
+                        <arg>-Ywarn-unused</arg> <!-- if you need exactly RemoveUnused -->
```
ohadbitt, line 361: what this means?

.scalafmt.conf
```diff
@@ -0,0 +1,34 @@
...
+runner.dialect = scala212
+fileOverride {
+  "glob:**/src/**/scala-2.13/**.scala" {
```
ohadbitt, line 30: this is not our folders structures

ohadbitt (approved):
As you said - i only looked at the configuration not the files - if you didn't change anything there - go ahead

.scalafmt.conf
ag-ramachandran: Thanks Ohad, fixed it to 120 chars.

connector/pom.xml
ag-ramachandran: If there are unused dependencies, this flag will report them. We eventually will add scalafix to fix code smells, we had to add this

.scalafmt.conf
ag-ramachandran: Fixed this

PR #401, opened 2024-09-19
Feature/update tag name for dups

ag-ramachandran:


connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -386,143 +365,152 @@ object KustoWriter {
...
+      // Need to copy the ingestionProperties so that only this blob ingestion will be effected
+      props = SparkIngestionProperties.cloneIngestionProperties(ingestionProperties)
     }
+    if (parameters.writeOptions.ensureNoDupBlobs) {
+      // The Key change is here
+      val tag = KDSU.getDedupTagsPrefix(parameters.writeOptions.requestId, s"${blobIndexInBatch.toString}_$partitionId")
+      KDSU.logInfo(
+        className,
+        s"With ensureNoDupBlobs in ${TaskContext.getPartitionId.toString} " +
+          s"for requestId: '${parameters.writeOptions.requestId}, the tag is $tag")
```
ohadbitt, line 465:
```suggestion
      val tag = KDSU.getDedupTagsPrefix(parameters.writeOptions.requestId, s"${blobIndexInBatch.toString}_$partitionId")
      KDSU.logInfo(
        className,
        s"With ensureNoDupBlobs in partition: ${TaskContext.getPartitionId}, " +
          s"for requestId: '${parameters.writeOptions.requestId}, tag: $tag")
```

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -386,143 +365,152 @@ object KustoWriter {
...
+
+  def ingest(
+      blobResource: BlobWriteResource,
+      parameters: KustoWriteResource,
+      ingestionProperties: IngestionProperties,
+      flushImmediately: Boolean = false,
+      blobUUID: String,
+      kustoClient: ExtendedKustoClient,
+      partitionsResults: CollectionAccumulator[PartitionResult],
+      batchIdForTracing: String,
```
ohadbitt, line 448: use the batchIdForTracing as well in getDedupTagsPrefix - so that it will work on (spark) streaming as well.

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -386,143 +365,152 @@ object KustoWriter {
...
+        blobIndexInBatch)
+    }
+  }
+
+  def ingest(
+      blobResource: BlobWriteResource,
+      parameters: KustoWriteResource,
+      ingestionProperties: IngestionProperties,
+      flushImmediately: Boolean = false,
+      blobUUID: String,
```
ohadbitt, line 445: blobUUID - not used

connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
```diff
@@ -386,143 +365,152 @@ object KustoWriter {
...
-        className,
-        s"Queued blob for ingestion in partition $partitionIdString " +
-          s"for requestId: '${parameters.writeOptions.requestId}")
-    }
     val kustoClient = KustoClientCache.getClient(
       parameters.coordinates.clusterUrl,
       parameters.authentication,
       parameters.coordinates.ingestionUrl,
       parameters.coordinates.clusterAlias)
+    val blobIdMap = new ConcurrentHashMap[String, Int]()
```
ohadbitt, line 377: did you mean for this to be some static cache ? if so it will need to live outside the function as a member.
currently this will always hold only the current run with the same request id in all Map entries, 
Regardless - a simple integer with ncrements would do the job i think

Issue #415, opened 2025-01-15
PySpark: simple read "take n" query fails with auth error if n larger than ~1MB


ViaFerrata:
**Describe the bug**
Below PySpark read query fails with an authentication (!) error if `n` in a `take n` command is 1e4 (and larger), 1e3 and below works fine.
The query limit of the DB is 5e6 and 1e4 rows is about 1 MB (the table contains openssh log sample data from [here](https://github.com/logpai/loghub/blob/master/OpenSSH/OpenSSH_2k.log_structured.csv), slightly adjusted).
The same query in the ADX web GUI works fine.

```python
df_kusto = spark.read.format('com.microsoft.kusto.spark.datasource'). \
            option("kustoCluster", cluster). \
            option("kustoDatabase", kusto_db). \
            option("kustoQuery", f'{table_name} | take 10000'). \  # querying whole table using only table_name also doesn't work
            option("kustoAadAppId", kusto_client_id). \
            option("kustoAadAppSecret", kusto_client_secret). \
            option("kustoAadAuthorityID", authority_id). \
            load()

print(df_kusto.count())

# Error message:
# Py4JJavaError: An error occurred while calling o454.count.
# shaded.databricks.org.apache.hadoop.fs.azure.AzureException: hadoop_azure_shaded.com.microsoft.azure.storage.StorageException: Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.

```

Running the code above again yields a different, more general spark error:
```
org.apache.spark.SparkException: [INTERNAL_ERROR] The Spark SQL phase planning failed with an internal error. You hit a bug in Spark or the Spark plugins you use. Please, report this bug to the corresponding communities or vendors, and provide the full stack trace. SQLSTATE: XX000
```

Can send the notebook containing the code if required.

**Expected behavior**
No failure.

**Files**

[stacktrace_1st_error.txt](https://github.com/user-attachments/files/18427555/stacktrace_1st_error.txt)
[stacktrace_2nd_error.txt](https://github.com/user-attachments/files/18519608/stacktrace_2nd_error.3.txt)

**Desktop (please complete the following information):**
 - OS: Azure Databricks, DBR 15.4 LTS, Standard DS3_v2, single node cluster
 - Connector: 3.0_2.12:5.2.4

Thanks a lot :)

ViaFerrata:
Update:
[...]

PR #417, opened 2025-01-27
Fix distributed mode

ohadbitt:


ag-ramachandran (approved):
Addresses : https://github.com/Azure/azure-kusto-spark/issues/415

Issue #267, opened 2022-11-19
Count query fails if a large query ends in a comment


oglehb:
**Describe the bug**
When the final line of a Kusto query ends with a comment, the count query fails if the tabular expression result contains more than `truncationmaxrecords` rows (500000 by default).

**To Reproduce**

    conn = {
        "kustoCluster": xxx,
        "kustoDatabase": xxx,
        "kustoAadAppId": xxx,
        "kustoAadAppSecret": xxx,
        "kustoAadAuthorityID": xxx,
    }
    crp = '{"Options": {"truncationmaxrecords": 500000}, "Parameters": {}}'
    kql = 'range _ from 1 to 500001 step 1 // this comment causes issue'

    df = (spark.read
            .format("com.microsoft.kusto.spark.synapse.datasource")
            .options(**conn)
            .option("kustoQuery", kql)
            .option("clientRequestPropertiesJson", crp)
            .load())
   
    df.limit(0)

**Expected behavior**
All else the same, I expected the query `range _ from 1 to 500001 step 1 // this comment causes issue` to behave identically to the query `range _ from 1 to 500001 step 1` (when `truncationmaxrecords` is 500000 or below in this case).

**Screenshots**

**Desktop (please complete the following information):**
 - Spark v3.2 on Synapse

**Additional context**
Maybe it has something to do with this code:
https://github.com/Azure/azure-kusto-spark/blob/83cf263a7e56365a52f895cf774fb24b72e9e0fc/connector/src/main/scala/com/microsoft/kusto/spark/utils/KustoQueryUtils.scala#L11-L13
Perhaps changing would solve the issue:

    def limitQuery(query: String, limit: Int): String = {
        val trimmedQuery = query.trim
        trimmedQuery + s"\n| take $limit"
    }

(I do not know Scala)

ag-ramachandran:
Hello @oglehb , Will have a look at this right away and potentially fixing this issue

monishkadas-ms:
The fix has been merged. Closing the issue now. Thank you!

PR #499, opened 2026-07-24
fix: prevent trailing // comment from swallowing appended query operators (#267)

monishkadas-ms:
A KQL // comment runs to end-of-line, but the connector appended its operators on the
same line as the user query — so a query ending in // comment silently swallowed them.

This fixes two failures:

Read fails on large results (#267): the schema probe | take 0 and the
| count / | evaluate estimate_rows_count() queries were absorbed into the comment,
so the raw query ran and tripped E_QUERY_RESULT_SET_TOO_LARGE past truncationmaxrecords.
Silent wrong data: filter/column pushdown | where / | project were also swallowed;
a dropped | project returned all columns and Spark mapped them positionally, so
df.select(col) could return another column's values.
Fix: append these operators after a newline (\n) instead of a space — semantically
identical KQL for normal queries, but it terminates any trailing comment. Sites changed:
KustoQueryUtils.limitQuery, CslCommandsGenerator.generateCountQuery /
generateEstimateRowsCountQuery, KustoFilter.buildFiltersClause / buildColumnsClause.

Future Release Comment
Fixes:

Read queries ending in a single-line comment (// ...) no longer disable the connector's
appended operators (| take, | count, | evaluate estimate_rows_count(), | where,
| project), fixing read failures on large results and a silent column-pruning corruption (#267)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ Tags are created as `v4.0_{version}` (Spark 4 / master) and `v3.0_{version}` (Sp
 - None
 
 ### Fixed
-- None
+- Read queries ending in a single-line comment (`// ...`) no longer disable the connector's appended operators. The generated `| take`, `| count`, `| evaluate estimate_rows_count()`, `| where`, and `| project` are now placed on a new line so a trailing comment cannot swallow them — fixing reads that failed with `E_QUERY_RESULT_SET_TOO_LARGE` on large results and a silent column-pruning corruption (#267)
 
 ## [7.1.1] - 2026-06-12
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -15,7 +15,7 @@ Tags are created as `v4.0_{version}` (Spark 4 / master) and `v3.0_{version}` (Sp
 - None
 
 ### Fixed
-- None
```
asaharn, line 18: For changelog, we can add entries at time of releasing a version also. Both ways are fine.

asaharn (approved):

PR #502, opened 2026-08-03
fix(ci): prevent Azure AD access token leaking into public workflow logs (CWE-532)

tanmaya-panda1:
Remediation for IcM 842305145 (Glasswing Mythos - July, follow-on wave to Twin Flames).

## Issue

`.github/workflows/build.yml` obtained a live Azure AD bearer token and propagated it unmasked:

```yaml
access_token=$(az account get-access-token ... --query accessToken -o tsv)
echo "ACCESS_TOKEN=$access_token" >> $GITHUB_ENV   # never masked
...
mvn clean verify ... -DaccessToken=${{env.ACCESS_TOKEN}}   # expanded into the printed command
```

GitHub Actions only auto-masks values that come from the `secrets.*` context, not values derived from them at runtime. The token was therefore never registered as a secret, and the runner prints both the resolved `run:` block and the step environment into the job log. Because this repository is public, any user with read access could download a run log and recover a live Azure AD bearer token for the test cluster.

## Fix

- Emit `::add-mask::` for the token before it is written to `$GITHUB_ENV`, so it is redacted anywhere it surfaces. Masking is not retroactive, so it has to precede the write.
- Drop `-DaccessToken=${{env.ACCESS_TOKEN}}` from the `mvn` command line. This was the value being expanded into the printed command, and it was redundant:
  - `KustoOptions.scala:36` defines the option name as the literal string `accessToken`
  - `KustoTestUtils.getSystemVariable` (`KustoTestUtils.scala:250-255`) reads `System.getenv` before falling back to `System.getProperty`
  - the step `env:` block already sets an environment variable named `accessToken`, which the forked test JVM inherits
  - `scalatest-maven-plugin` (`connector/pom.xml:268-276`) declares an `<argLine>` but no `<systemProperties>`, so Maven user properties were not forwarded into the fork anyway

The `env: accessToken` entry is deliberately retained, because not every E2E test is covered by `<tagsToExclude>KustoE2E</tagsToExclude>` (for example `KustoSourceE2E.scala:130,190,197,237`), so those tests still need the value at runtime.

- The same pattern existed in `.github/workflows/release.yml`, which was not flagged by the scan. Fixed there too.

## Validation

- Both workflow files parse as valid YAML.
- No remaining unmasked path from the token to a logged command.
- `.azuredevops/build.yml` and `.azuredevops/release.yml` were checked and are 3-line placeholders with no equivalent issue.

## Follow-up, not in this PR

- The credential behind `secrets.APP_ID` should be rotated, since this fix prevents future exposure but does not undo past exposure.
- `actions/checkout` persists GitHub credentials by default; `persist-credentials: false` is worth considering for these privileged jobs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,6 +37,9 @@ jobs:
       - name: "Run az commands"
         run: |
               access_token=$(az account get-access-token --resource=${{ secrets.APP_ID }} --scope=${{ secrets.CLUSTER }}/.default --query accessToken -o tsv)
+              # Register the derived token as a secret so it is redacted from logs.
+              # GitHub only auto-masks values that come from secrets.*, not values derived from them.
+              echo "::add-mask::$access_token"
               echo "ACCESS_TOKEN=$access_token" >> $GITHUB_ENV
       - name: Run the Maven verify phase
         timeout-minutes: 40
@@ -50,7 +53,7 @@ jobs:
           ingestStorageUrl: ${{ secrets.INGEST_STORAGE_URL }}
           ingestStorageContainer: ${{ secrets.INGEST_STORAGE_CONTAINER }}
         run: |
-          mvn clean verify -B -DkustoAadAppId=${{ secrets.APP_ID }} -DkustoAadAuthorityID=${{ secrets.TENANT_ID }} -DkustoDatabase=${{ secrets.DATABASE }} -DkustoCluster=${{ secrets.CLUSTER }} -DaccessToken=${{env.ACCESS_TOKEN}} -DingestStorageUrl=${{secrets.INGEST_STORAGE_URL}} -DingestStorageContainer=${{secrets.INGEST_STORAGE_CONTAINER}}
+          mvn clean verify -B -DkustoAadAppId=${{ secrets.APP_ID }} -DkustoAadAuthorityID=${{ secrets.TENANT_ID }} -DkustoDatabase=${{ secrets.DATABASE }} -DkustoCluster=${{ secrets.CLUSTER }} -DingestStorageUrl=${{secrets.INGEST_STORAGE_URL}} -DingestStorageContainer=${{secrets.INGEST_STORAGE_CONTAINER}}
           
       - name: Publish Unit Test Results
         uses: EnricoMi/publish-unit-test-result-action@v2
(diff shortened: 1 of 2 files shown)


tanmaya-panda1:
CI note for reviewers: this commit produced **two** GitHub Actions runs (the workflow triggers on both `push` and `pull_request`), and they disagree despite being the exact same code at `3086c82`:

| Run | Trigger | Result |
|---|---|---|
| [30790277875](https://github.com/Azure/azure-kusto-spark/actions/runs/30790277875) | `push` | **success** - `Run the Maven verify phase => success` |
| [30790316017](https://github.com/Azure/azure-kusto-spark/actions/runs/30790316017) | `pull_request` | failure |

The two runs started one minute apart and overlapped, executing the E2E suite concurrently against the same shared Kusto cluster and the same transient storage. The failing run shows the classic symptoms of that contention rather than a code fault:

- `Account sparkwesteu does not exist`
- `SEM0100: 'count' operator: Failed to resolve table or column expression named 'KustoBatchSinkE2E_Ingest_DMStorage_...'` - the count runs against a table whose ingestion did not land

Authentication demonstrably worked in the failing run too: it refreshed ingestion resources and received a Kusto *semantic* error, which only a successfully authenticated and authorized session can reach. A broken token would surface as 401/Unauthorized instead.

The passing run confirms the change is correct end to end, which is the important part given this PR removes `-DaccessToken` from the Maven command line:

```
Run az commands            echo "::add-mask::$access_token"
Run the Maven verify phase   accessToken: ***
Run the Maven verify phase => success
```

So on the fixed workflow: the mask is registered, the environment value the tests actually consume is now redacted as `***` rather than printed in clear text, no raw JWT appears anywhere in the log, and the full E2E suite still passes without the removed system property. That is exactly the behaviour this PR intends.

Worth noting separately: `accessToken: ***` appearing in the log confirms the runner does print the step `env:` block, so masking - not just removing the value from the command line - was required here.

The two ADO checks (`Azure-Kusto-Spark-Build`, `Azure-Kusto-Spark-Release`) also fail identically on already-merged PRs such as #501, and correspond to the comment-only placeholder files in `.azuredevops/`. They are unrelated to this change.

Re-running the PR-triggered job on its own, without a concurrent run, should make it green.

PR #503, opened 2026-08-03
fix(ci): stop materializing the Azure AD access token in CI (CWE-532)

tanmaya-panda1:
Follow-up to #502 (stacked on it — please merge #502 first).

## Problem

#502 stopped the token leaking into public logs by registering it with `::add-mask::`. That closes the disclosure, but the token is still **minted as a workflow-level value**: the `Run az commands` step wrote a live Azure AD bearer token to `$GITHUB_ENV`, and it was then injected into the Maven step's `env:` block.

Masking is a redaction control, not containment. On a public repository the safer property is for the credential to never become a workflow value at all.

## Change

Delete the `Run az commands` step and the `accessToken:` env var from `build.yml` and `release.yml`. Nothing added — 8 lines removed from each file.

`azure/login@v2` is retained.

## Why this is safe

`KustoTestUtils.getKustoConnectionOptions` already has exactly this fallback:

```scala
val maybeAccessTokenEnv = Option(getSystemVariable(KustoSinkOptions.KUSTO_ACCESS_TOKEN))
val accessToken: String = maybeAccessTokenEnv match {
  case Some(at) => at
  case None     => new AzureCliCredentialBuilder().build().getToken(tokenRequestContext).block()
}
```
(`KustoTestUtils.scala:208-233`)

With the variable unset, the credential chain acquires the token **inside the JVM** from the CLI session `azure/login@v2` established. It is never a workflow value, so there is nothing to mask.

The prerequisite is already proven in CI: `KustoSourceE2E.scala:199` calls `generateSasDelegationWithAzCli`, which shells out to `az` and depends on that same authenticated session.

Note this also required removing the `env:` line, not just the step — `getSystemVariable` returns the env var if present, and `accessToken: ${{env.ACCESS_TOKEN}}` with `ACCESS_TOKEN` unset yields an **empty string**, which `Option("")` would wrongly treat as `Some("")`.

## Verified

- `KUSTO_ACCESS_TOKEN` is literally `"accessToken"` (`KustoOptions.scala:36`) — matches the removed env var
- No other consumer of that env var or system property exists in `connector/` or `samples/`
- No residual `ACCESS_TOKEN` reference in either workflow
- Both workflows still parse as valid YAML
- The scope is now correctly `https://<cluster>.kusto.windows.net/.default` (via `clusterToKustoFQDN`), the standard Kusto AAD scope

`accessToken` remains a fully supported connector option — only CI stops supplying it.

## Consistency

This matches the pattern in `azure-kusto-python`, `azure-kusto-java`, `azure-kusto-go` and `azure-kusto-node`, which authenticate with OIDC and let the SDK credential chain acquire tokens in-process. `azure-kusto-spark` was one of only two Kusto repos that materialized a raw token.

Related: IcM 842305145 (Glasswing Mythos – July).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,13 +34,6 @@ jobs:
           client-id: ${{ secrets.APP_ID }}
           tenant-id: ${{ secrets.TENANT_ID }}
           subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
-      - name: "Run az commands"
-        run: |
-              access_token=$(az account get-access-token --resource=${{ secrets.APP_ID }} --scope=${{ secrets.CLUSTER }}/.default --query accessToken -o tsv)
-              # Register the derived token as a secret so it is redacted from logs.
-              # GitHub only auto-masks values that come from secrets.*, not values derived from them.
-              echo "::add-mask::$access_token"
-              echo "ACCESS_TOKEN=$access_token" >> $GITHUB_ENV
       - name: Run the Maven verify phase
         timeout-minutes: 40
         env:
@@ -48,7 +41,6 @@ jobs:
           kustoDatabase: ${{ secrets.DATABASE }}
           kustoCluster: ${{ secrets.CLUSTER }}
           kustoAadAppId: ${{secrets.APP_ID}}
-          accessToken: ${{env.ACCESS_TOKEN}}
           storageAccountUrl: '${{secrets.STORAGE_CONTAINER_URL}}'
           ingestStorageUrl: ${{ secrets.INGEST_STORAGE_URL }}
           ingestStorageContainer: ${{ secrets.INGEST_STORAGE_CONTAINER }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,6 +37,9 @@ jobs:
       - name: "Run az commands"
         run: |
               access_token=$(az account get-access-token --resource=${{ secrets.APP_ID }} --scope=${{ secrets.CLUSTER }}/.default --query accessToken -o tsv)
+              # Register the derived token as a secret so it is redacted from logs.
+              # GitHub only auto-masks values that come from secrets.*, not values derived from them.
+              echo "::add-mask::$access_token"
               echo "ACCESS_TOKEN=$access_token" >> $GITHUB_ENV
       - name: Run the Maven verify phase
         timeout-minutes: 40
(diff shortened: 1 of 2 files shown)


tanmaya-panda1:
## ✅ Validated end-to-end in CI

The concern with this change was whether the `AzureCliCredentialBuilder` fallback would actually acquire a working token once the workflow stopped supplying `accessToken`. Both runs on this branch confirm it does.

**Run [30794649666](https://github.com/Azure/azure-kusto-spark/actions/runs/30794649666) (push) and [30794674606](https://github.com/Azure/azure-kusto-spark/actions/runs/30794674606) (pull_request) — both `success`.**

```
Total number of tests run: 166
Tests: succeeded 166, failed 0, canceled 0, ignored 0, pending 0
All tests passed.
[INFO] BUILD SUCCESS
```

### The fallback demonstrably took the intended branch

From the run log:

| Log marker | Count | Meaning |
|---|---|---|
| `Using scope <...> and authority <...>` | 1 | token-acquisition path executed |
| `Using access token from environment variable` | **0** | env var was *not* used → `AzureCliCredentialBuilder` fallback taken |

So the token was acquired **inside the JVM** from the CLI session established by `azure/login@v2`, exactly as intended — not read from the workflow environment.

### E2E suites that actually ran

`KustoAuthenticationTestE2E`, `KustoSourceE2E`, `KustoBatchSinkE2E`, `KustoSinkBatchE2E`, `KustoSinkStreamingE2E`, `KustoSinkSchemaAdjustmentE2E`, `KustoBlobAccessE2E`, `KustoPruneAndFilterE2E`

`KustoAuthenticationTestE2E` passing is the direct confirmation that authentication is healthy, and `KustoSourceE2E` exercises `generateSasDelegationWithAzCli`, which independently depends on the same CLI session.

### No credential in the log

Scanned the full 910 KB log (counts only, values never printed):

| Check | Count |
|---|---|
| JWT-shaped strings (`eyJ….…`) | **0** |
| `accessToken:` env references | **0** |
| `az account get-access-token` invocations | **0** |

The credential no longer appears anywhere in the workflow.

### Regression review

The change is purely subtractive — 8 lines removed per workflow, nothing added.

- `KUSTO_ACCESS_TOKEN` is literally `"accessToken"` (`KustoOptions.scala:36`), matching the removed variable
- No other consumer of that env var or system property exists in `connector/` or `samples/`
- Removing the `env:` line was **required**, not incidental: with the step gone but the line kept, `accessToken: ${{env.ACCESS_TOKEN}}` resolves to an **empty string**, and `Option("")` returns `Some("")` — which would have silently bypassed the fallback and failed auth with a blank token
- Both workflows still parse as valid YAML

### Unrelated check failures

`Azure-Kusto-Spark-Build` / `Azure-Kusto-Spark-Release` (ADO) fail identically on #502 and on already-merged #501. `.azuredevops/build.yml` and `release.yml` are 140-byte comment-only placeholders with no pipeline definition, so those checks cannot pass on any branch. Not related to this change.

PR #510, opened 2026-09-03
fix: prevent CloudInfo executor stalls 

monishkadas-ms:
## Problem

Queued and Transactional writes could stall indefinitely when Kusto SDK 5.1.1 performed repeated executor-side `CloudInfo` metadata requests under a JVM-wide lock. A single non-responsive request could block every Spark executor slot.

## Fix

- Retrieve ingestion metadata once on the driver with a 40-second caller deadline.
- Perform metadata preflight before Kusto resource mutations or Spark partition execution.
- Pass the resulting `CloudInfo` to executors.
- Seed each executor cache through the same bounded helper before evaluating the lazy queued-ingestion client.
- Discover the runtime cache API reflectively, supporting both `(String, CloudInfo)` and `(String, Mono<CloudInfo>)` without a direct cache-method or Reactor descriptor in `KustoWriter` bytecode.
- Preserve existing Streaming behavior, SDK exception classification, and the five-field `KustoWriteResource` shape.

## Validation

- Java 8 focused tests: **17/17 passed**.
- Clean shaded package: **SUCCESS**.
- Databricks validation passed all 19 mandatory gates and both optional write gates.

## Known limitation

The caller deadline does not terminate SDK 5.1.1’s underlying Apache socket read. A true metadata blackhole can leave that JVM’s SDK cache monitor/helper poisoned, so later callers fail at their own deadlines and the driver must be restarted.

This hotfix prevents Spark executor-slot starvation; it does not make the affected JVM self-healing.

The private Synapse wrapper/VHD must also verify that driver and executors load the same `CloudInfo` class version before production rollout.

## Scope

Limited to the executor-liveness mitigation. It does not include the separate queue-timeout unit correction, an SDK upgrade, or unrelated dependency changes.

---

#### Future Release Comment

**Breaking Changes:**
- None

**Features:**
- None

**Fixes:**
- Prevent Queued and Transactional writes from indefinitely consuming Spark executor slots when Kusto metadata retrieval or cache insertion stalls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,17 @@ Tags are created as `v4.0_{version}` (Spark 4 / master) and `v3.0_{version}` (Sp
 ### Fixed
 - None
 
+## [5.3.6] - 2026-09-01
+
+### Changed
+- None
+
+### Added
+- None
+
+### Fixed
+- Prevent queued and transactional writes from blocking Spark executors during ingestion metadata retrieval
+
 ## [5.3.5] - 2026-06-02
 
 ### Changed
--- a/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
+++ b/connector/src/main/scala/com/microsoft/kusto/spark/datasink/KustoWriter.scala
@@ -48,6 +48,7 @@ import org.apache.spark.sql.types.StructType
 import org.apache.spark.util.CollectionAccumulator
 
 import java.io._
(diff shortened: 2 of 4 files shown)


tanmaya-panda1:
Please validate these scenarios as well

- multiple tasks can initialize together which may lead to blocked-cache callers time out, validate that cancelling one caller does not affect others.

- the deployed JAR successfully transfers CloudInfo and populates the SDK cache on executors.

tanmaya-panda1 (approved):

